Issue #49, opened 2016-03-01
Add Sections to a Monograph


mark-dce:
**STORY**
As a press user, I want to be able to create sections within a monograph, so that I can model the chapters and other subdivisions in a monograph.

**ACCEPTANCE**
- [x] I can create sections corresponding to the chapters in a monograph
- [x] I can create sections corresponding to appendices and other non-chapter divisions
- [x] Section titles appear on the the monograph show page

PR #58, opened 2016-03-10
Display sections on monograph show page. Part of story #49

val99erie:


val99erie:
@jcoyne Please review

app/presenters/monograph_presenter.rb
```diff
@@ -0,0 +1,3 @@
+class MonographPresenter < ::CurationConcerns::WorkShowPresenter
+  delegate :member_docs, :section_docs, to: :solr_document
```
jcoyne: Why are we delegating to solr_doc instead of just adding that here?  Is there another scenario where it will be used?

app/views/curation_concerns/monographs/_sections.html.erb
```diff
@@ -0,0 +1,11 @@
+<% unless @presenter.section_docs.blank? %>
```
jcoyne: Use the passed in local: `presenter` rather than the ivar `@presenter`.

app/presenters/monograph_presenter.rb
```diff
@@ -0,0 +1,16 @@
+class MonographPresenter < ::CurationConcerns::WorkShowPresenter
+  def member_docs
```
jcoyne: Should this be a private method?

PR #61, opened 2016-03-11
Add Sections to a Monograph

### LINKED ISSUE #49: Add Sections to a Monograph

jcoyne:
Fixes #49


spec/presenters/curation_concerns/section_presenter_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      .with("{!terms f=member_ids_ssim}qr46r109z")
+      .and_return(response)
+  end
+
+  describe "#monograph_label" do
+    subject { presenter.monograph_label }
+    it { is_expected.to eq 'My first monograph' }
+  end
+
+  # describe "#monograph_id" do
```
val99erie: From what I can tell, monograph_id isn't stored on the fedora object for the Section, is that right?  How will it work for editing an existing section?  How will the monograph_id be populated on the form?

jcoyne: monograph_id will just be fetched from solr. This becomes necessary if we want to link from the show page of the section back to the monograph.

PR #63, opened 2016-03-11
Display fields on Monographs show page

### LINKED ISSUE #53: Implement the initial metadata fields for a Monograph
**DETAILS**
This is the implementation story that correlates with #8 

**ACCEPTANCE**
- [x] Any new fields are implemented in the metadata code.
- [ ] Fields are displayed on the Monograph show page
- [x] Field contents can be edited in the Monograph edit view.

jmcglone:
Closes #53. Modifies presenter class for Monographs, update the solr document, updates the Monograph controller, overrides the curation_concerns view partial and uses snake_case for all instances of buy_url.


app/controllers/curation_concerns/monographs_controller.rb
```diff
@@ -1,5 +1,10 @@
 class CurationConcerns::MonographsController < ApplicationController
   include CurationConcerns::CurationConcernController
   set_curation_concern_type Monograph
+  # Use the Monograph Presenter
+  def show_presenter
```
jcoyne: This is already set on line 9 (new way), so there's no need to include these lines.

app/models/curation_concerns/_form_additional_information.html.erb
```diff
@@ -0,0 +1,12 @@
+<fieldset class="optional prompt">
+  <legend>Additional Information</legend>
+  <%= f.input :subject,                as: :multi_value, input_html: { class: 'form-control' } %>
+  <%= f.input :publisher,              as: :multi_value, input_html: { class: 'form-control' } %>
+  <%= f.input :date %>
+  <%= f.input :source,                 as: :multi_value, input_html: { class: 'form-control' } %>
+  <%= f.input :language,               as: :multi_value, input_html: { class: 'form-control' } %>
+  <%= f.input :isbn,                   as: :multi_value, input_html: { class: 'form-control' }, label: 'ISBN' %>
```
jcoyne: You may want to consider putting these labels in the i18n. the key would be `en.simple_form.labels.defaults.isbn` for this one. If you don't think that's required, I don't think we need to hold up the PR though.

jmcglone: This is a great suggestion. I'll get with Melissa and Jeremy re: requirements and if it is we can add as a story for 0.2 release sprint.

jcoyne: FWIW, I wouldn't call it a story. A story must be something that's visible to the users, this wouldn't be.  This is more "keeping everything organized" so you know what file to open to change it.

app/controllers/curation_concerns/monographs_controller.rb
jmcglone: This has been removed in an amended commit and push.

jcoyne: I don't think push went through. Did you do a force push (`--force`)?

jmcglone: I did to a force push, but I'm on my work machine now with this branched merged locally. I'll close this PR and issue a new one. Sorry.

app/models/curation_concerns/_form_additional_information.html.erb
jmcglone: Ok, I've added these labels to the locales/en.yml file.

jcoyne: Good. It may also be good to remove the `label` option here, because that will override any changes to the i18n.

jmcglone: I'm not sure why this isn't showing up in the diff, but they were removed in commit 77171e6

jcoyne: I'll bet it's not showing the full, diff, only the last commit.

PR #75, opened 2016-03-16
Editor can publish a book and all its components

jcoyne:


app/jobs/publish_job.rb
```diff
@@ -0,0 +1,17 @@
+class PublishJob < ActiveJob::Base
+  queue_as :publish
+
+  def perform(curation_concern)
+    curation_concern.date_published = [CurationConcerns::TimeService.time_in_utc]
+
+    # TODO: this is probably not how we publish in the long run
+    curation_concern.visibility = 'open'
```
val99erie: Is there a contant we can use instead of hardcoding 'open', such as Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC

PR #102, opened 2016-03-17
Generate a press and role model

### LINKED ISSUE #97: Manage list of presses
**STORY**
As a platform admin, I need a mechanism to add new presses, so that I can grow our hosting business.

**ACCEPTANCE**
For each press, the system maintains a record of
- [ ] The press name
- [ ] A pointer to the press logo
- [ ] A description of the press
- [ ] The subdomain for the press (ID or short-name)

### LINKED ISSUE #100: Create a platform admin
**STORY**
I'd like to be able to create platform-wide admin to do other setup activities.

**DETAILS**
Maybe have this done as part of the bin/setup script - i.e. prompt for the e-mail and password for a first user.  Create that user and grant them all admin privileges in the system.

**ACCEPTANCE**
- [ ] After setting up the application, I can login with the platform admin account.
- [ ] The user has the platform_admin role

jcoyne:
Fixes #97 Fixes #100

### DIFF
--- a/lib/tasks/roles.rake
+++ b/lib/tasks/roles.rake
@@ -1,0 +1,49 @@
+desc 'Add application-wide admin privileges to a user'
+task admin: :environment do
+  puts 'Creating an admin user.'
+  u = prompt_to_create_user
+  Role.create(user: u, resource: nil, role: 'admin')
+  puts 'User created.'
+end
+
+desc 'Create a new press'
+task press: :environment do
+  print 'Press name: '
+  name = $stdin.gets.chomp
+
+  press = Press.create!(name: name)
+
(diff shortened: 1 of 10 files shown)


lib/tasks/roles.rake
```diff
@@ -0,0 +1,49 @@
+desc 'Add application-wide admin privileges to a user'
+task admin: :environment do
+  puts 'Creating an admin user.'
+  u = prompt_to_create_user
+  Role.create(user: u, resource: nil, role: 'admin')
+  puts 'User created.'
+end
+
```
val99erie: Could you add some info to the README so that people know they can use these rake tasks to created admin users and presses.

Issue #50, opened 2016-03-01
Order Sections within a Monograph


mark-dce:
**STORY**
As a press user, I would like to be able to order sections within a monograph, so that I can reflect the ordering and structure of the primary narrative in the UI for accessing associated assets.

**NOTES**
For the first pass, do the simplest possible thing.  This will facilitate having a more detailed discussion about refinements like drag-and-drop vs. assigning order numbers, etc.

**ACCEPTANCE**
- [x] I can re-arrange the order of sections within a Web UI

val99erie:
If you click the "File Manager" button, you can reorder all the sections and assets that are attached to the monograph by using the drag and drop functionality.

val99erie:
Now that the user roles are in, you can't just create monographs anymore.  Here's what you'll need to do to test this story:

Run this rake task to create the super admin user:
rake admin

And make sure redis is running.

Then you can log into the UI with that user and create a monograph with sections and files attached.  At the bottom of the monograph show page, click "File Manager" (we should change the name of that button), and you should be able to reorder the files and sections.

The re-ordering won't take effect unless you click the 'Save' button.

Then when you go back to the monograph show page, the files and sections are each listed separately, but each list should be in the correct order.

We'll probably change how these lists are displayed at some point (like maybe they won't be in separate lists), but this was just a first pass.

PR #111, opened 2016-03-22
Order sections within a monograph. Story #50

val99erie:


jmcglone:
@acozine when I run `bundle exec bin/setup` I'm seeing this deprecation warning:

`DEPRECATION WARNING: ActiveFedora::SolrQueryBuilder.solr_name is deprecated and will be removed in ActiveFedora 10.0. Use ActiveFedora.index_field_mapper.solr_name instead.`

I don't know DCE's practice and whether these kinds of things should hold up a PR, so I'll let you make the call.

val99erie:
@jmcglone We shouldn't let the deprecation warnings hold up the PR.  They weren't introduced in this PR, and the warnings are coming from the underlying hydra gem.  There are at least 2 gems that we'll have to update before we can get rid of that message.

jmcglone:
@val99erie Got it! Ok, another question. I've created a monograph with sections, but I can't see where to re-order the sections in the UI.

val99erie:
Ah, Justin added the role stuff in, so the UI has changed.  There used to
be a "Manage Files" button, or something like that at the bottom of the
monograph show page.  Then you can re-order things in there.  But you might
need to make your user an admin first.  I'll try to figure out the steps
needed to do that.

Valerie

Issue #95, opened 2016-03-17
Manually associate a new monograph with a press


mark-dce:
**STORY**
When I create a new monograph, I would like to be able to associate it with a press, so that it will show up on that press's landing page and facets.

**ACCEPTANCE**
- [x] Users get a list of presses to choose from when creating a monograph (see #97 re list of presses)
- [x] A monograph must be associated with a press
- [x] Update the dev notes description of how to create a monograph:  https://github.com/curationexperts/heliotrope/wiki/Developer-Notes

mark-dce:
Blocked until #97 is complete

val99erie:
@mark-dce Some assumptions I made...

For the predicate of the Press, I used:
http://id.loc.gov/vocabulary/relators/pbl

And curation_concerns already has a predicate for Publisher, which is a multi-value field:
http://purl.org/dc/elements/1.1/publisher

So, I added the new predicate for press, but I didn't change the existing publisher property.
If you want something different, could you write a story with the new predicate info?

PR #130, opened 2016-03-30
Associate a monograph with a press. Story #95

val99erie:


jmcglone:
This looks great. Thanks Val!

@val99erie @jgmorse @mark-dce For future discussion: When dealing with Monograph metadata, what's the difference between a Publisher and a Press? Will there ever be a case where the Publisher is different than the Press? If we don't think so, by default the publisher field should inherit the value of the Press with the ability to overwrite it if needed, or, if don't anticipate the need for a custom publisher, maybe the Press field should just replace the publisher field and take on its existing and planned behavior (be facet-able across the entire platform, searchable etc.).

PR #137, opened 2016-04-04
Use yaml file for curation_concerns server settings

acozine:
If this seems like a good approach, I will make the changes upstream in curation_concerns itself and open a PR there.


mark-dce:
Looks good.  I wonder if it would be clearer to call the configuration file `curation_concerns_paths.yml.sample` or something similar instead of `cc.yml.sample`.  It may not be obvious to a newcomer that "cc" stands for "curation_concerns" and not "carbon_copy" or something else.  I.E.  I have to open that file and read it to figure out what it relates to because "cc" is pretty brief and I might not make the immediate connection to "curation_concerns"

mark-dce:
Looks good, merging

Issue #91, opened 2016-03-16
Minter statefile should not go in /tmp


jcoyne:
Currently its at /tmp/minter-state, but it belongs in a location that will survive a restart.

acozine:
Can close when https://github.com/curationexperts/ansible-hydra/pull/118 is merged.

mark-dce:
Looks like the related PR: curationexperts/ansible-hydra#118  is closed.

Can we deal with #148 as part of this ticket - I think #148 needs to be closed and re-envisioned:

Right now it looks like the default that development environments inherit a setting that points to the system /tmp directory, which gets erased every time you reboot your system.  It would be much better to persist the minter state in development environments too by putting it in the RAILS_ROOT/tmp directory.

PR #148, opened 2016-04-13
Persist minter-state for development enviornments

mark-dce:
Moving minter-state from system /tmp which gets erased every boot to project ./tmp subdirectory so minter-state will persist in dev environments.


acozine:
Changing our approach - see #91.

PR #168, opened 2016-04-18
Persist minter-state for development environments

mark-dce:
Moves minter-state from system /tmp which gets erased every boot
to project tmp subdirectory so minter-state will persist in development.

Also provides safer defaults for production settings for folks
who deploy directly from the repo without using capistrano.


acozine:
Looks good. I wouldn't recommend using $RAILS_ROOT/tmp/ as the home for derivatives or the minter-state file in production, but the best place to set an alternative location is in the production section of the yaml file.

Issue #165, opened 2016-04-15
Platform admin can manage monographs created by other users.


val99erie:
Background info about the intended user abilities for heliotrope:
https://docs.google.com/spreadsheets/d/1D-tFBPzW50InN3qPnZzNScihgbsIQO87DO0mWfNlptg/edit#gid=1442564538

ACCEPTANCE CRITERIA:
- [x] A platform-level admin user can read a monograph created by another user
- [x] A platform-level admin user can update a monograph created by another user
- [x] A platform-level admin user can destroy a monograph created by another user
- [x] A platform-level admin user can publish a monograph created by another user
- [x] A normal logged in user (non-admin) **cannot** update a monograph created by another user
- [x] A normal logged in user (non-admin) **cannot** destroy a monograph created by another user
- [x] A normal logged in user (non-admin) **cannot** publish a monograph created by another user

Note: For a non-admin user, we'll leave the read scenario for another story, since it might be complicated.  It will depend on visibility & publish status, which we haven't fully defined yet.

val99erie:
Check out: `spec/models/ability_spec.rb` and `app/models/ability.rb`

sethaj:
I've added some specs for the first 4 ACs, they pass.

For the final 3 a "normal logged in user (non-admin)" can't, right now, create/publish/destroy _at all_, either their own monographs or others. And there are already tests for that. So maybe this is done? Or we need to think about what a "normal logged in user (non-admin)" role is, like an Editor or Author maybe?

jmcglone:
I think the definition of a normal logged in user (non-admin) at this point is a general user who signs in. They should not be able to CUD. If there are specs for this already, I would consider this done.

When we get around to dealing with Users and Roles, and actually adding the Editor and Author roles to the app, yes we'll need more specs to test.

val99erie:
@sethaj Ah, yeah, I didn't realize that there were already tests for some of those cases.  There are already tests for update and destroy here.  Maybe you could add a line for publish, just for completeness:

https://github.com/curationexperts/heliotrope/blob/master/spec/models/ability_spec.rb#L124-L125

sethaj:
PR for this is #169

PR #169, opened 2016-04-19
wrote some tests for #165, platform admins can manage monographs crea…

sethaj:
…ted by others. 

PR for #165

### DIFF
--- a/spec/models/ability_spec.rb
+++ b/spec/models/ability_spec.rb
@@ -9,10 +9,12 @@ describe Ability do
   let(:section) { create(:section, user: creating_user) }
   let(:file_set) { create(:file_set, user: creating_user) }
 
-  describe 'platform-wide admin user' do
+  describe 'a platform-wide admin user' do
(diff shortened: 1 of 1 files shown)


spec/models/ability_spec.rb
```diff
@@ -108,10 +117,12 @@
         should_not be_able_to(:read, section)
         should_not be_able_to(:update, section)
         should_not be_able_to(:destroy, section)
+        should_not be_able_to(:publish, section)
 
         should_not be_able_to(:read, file_set)
         should_not be_able_to(:update, file_set)
         should_not be_able_to(:destroy, file_set)
+        should_not be_able_to(:publish, file_set)
```
val99erie: We shouldn't test "publish" for file_set or section.  The publish action is only available for monographs:
https://github.com/curationexperts/heliotrope/blob/master/config/routes.rb#L18-L22

sethaj: Good point, thanks. Let me update the PR.

Issue #207, opened 2016-04-29
Handle missing-press errors gracefully


acozine:
## STORY:

As user, when I hit a URL that doesn't exist, I shouldn't think the site is broken. We're currently returning 500 errors for presses that doesn't exist (press.curationexperts.com/michigan-state) and for the resque admin pane (press.curationexperts.com/resque) if the user is non-privileged.
### CURRENT BEHAVIOR:

User gets a 500 error (We're sorry, something went wrong). 
### DESIRED BEHAVIOR:

User gets a flash message (The press "michigan-state" doesn't exist) and gets redirected to the home page.

acozine:
Related to #197.

val99erie:
This can probably be fixed here, by adding a check for existence of the press before trying to load it:
https://github.com/curationexperts/heliotrope/blob/master/app/controllers/press_catalog_controller.rb#L22

jmcglone:
The only case this could be a problem is when a user types in the press URL which is printed in the book front matter. It would be good to test this during Winter 2017 usability testing (#719). 

An option could be to redirect a user to a page that lists all the presses if they type in a press URL that does not exist.

gkostin1966:
Implemented redirect_to(presses_path) option as recommended by @jmcglone.

Issue #197, opened 2016-04-28
Implement resque-web mounted in application


mark-dce:
**STORY**
As a platform or press administrator, I would like an easy way to view background jobs, so that I can verify queues are being processed correctly.

**DETAILS**
See the instructions for mounting resque-web as an engine here: https://github.com/resque/resque-web#starting

Adapt the instructions for restricting access here: https://github.com/resque/resque-web#starting
We'll need to modify the code to match our authorization scheme for admins.

**ACCEPTANCE**
- [x] As a platform admin, I can view the resque-web interface from my browser at \<host\>/resque
- [ ] ~~As a press admin, I can view the rescue-web interface from my browser~~
- [ ] ~~As a press editor, I get an error message saying the page does not exist~~
- [x] As any other logged in user, I get an error message saying the page does not exist
- [x] As an unauthenticated user, I get a message asking me to log in.
- [x] As a platform admin, I have a link to the resque-web URL

acozine:
First acceptance met by PR #208. I'm taking my name off this ticket and tossing it back into Ready.

val99erie:
@mark-dce Should a press-level admin really have access to resque?  That will allow them to view/delete/cancel/re-run jobs at other presses, and may expose data from other presses.

mark-dce:
At some point, the Michigan folks will need to decide if press-level admins should have full access, some kind of constrained access, or no access.  Ideally press-level admins would have access to jobs for their press, but not other presses, but I think it's a pretty big piece of work to make that work (you might need individual queues per press or to somehow namespace jobs by press, etc.)

It's 50/50 to me whether press-level administrators get the same access as platform-level administrators or whether they're restricted the same way press editors are.  I put the requirement in to ensure we addressed the access one way or another, but I'm totally ok if denying access is simpler.

jmcglone:
I agree it would be great if a press-level admin could have access to just
their jobs -- but for now I think it would be okay to deny access.

Curious if the kind of functionality Mark describes would be something
hy-box folks tackle with multi-tenancy?

On Thu, May 5, 2016 at 9:47 PM, Mark Bussey <notifications@github.com>
wrote:

> At some point, the Michigan folks will need to decide if press-level
> admins should have full access, some kind of constrained access, or no
> access. Ideally press-level admins would have access to jobs for their
> press, but not other presses, but I think it's a pretty big piece of work
> to make that work (you might need individual queues per press or to somehow
> namespace jobs by press, etc.)
>
[...]

PR #208, opened 2016-04-29
mount resque-web for platform admins

acozine:
Addresses the first acceptance criterion of #197. Only platform admins can access the resque-web console at /resque; other users currently get 500 errors. See also #207.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,8 @@ gem 'rails', '4.2.6'
 
 gem 'curation_concerns', '0.12.0'
 gem 'active-fedora', '9.10.4'
+gem 'resque-web', '~> 0.0.7', require: 'resque_web'
+gem 'resque'
 # gem 'pg', '0.18.4'
 gem 'mysql2'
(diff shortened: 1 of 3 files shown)


val99erie:
It seems to be working. I can visit the */resque URL and see the job queues, and when I upload a new file, I can see jobs get queued.

However, when I run resque-web from the command line:
`resque-web -F -N curation_concerns:development`
I can't see the same jobs/queues as in the mounted resque-web.

The mounted resque-web has this info at the bottom of the page:
`Connected to Redis namespace curation_concerns:development on redis://localhost:6379/0`

Is my command line incorrect?

val99erie:
Maybe it doesn't matter.  I don't think I'll need to run resque-web outside of the rails server, so we shouldn't let that question prevent us from merging the PR.

@acozine Could you update the wiki page to explain that resque-web is mounted onto the rails app, and the URL to get to it.  https://github.com/curationexperts/heliotrope/wiki/Background-Jobs

PR #209, opened 2016-05-02
add the monograph's id to its section(s) for #187

### LINKED ISSUE #187: Monograph should be persistent when editing a section
**STORY**
As a user, when I click on a Section to edit it, the "monograph" field / drop down should retain the name of the monograph the section is associated with. Currently, this field clears out. 

**ACCEPTANCE CRITERIA**
- [x] When I click on the Edit button within a Section, the Edit page loads with the correct monograph already populated. 

- [x] When I click on the Edit button with a Section, Monograph field is not blank/empty when the Edit pages loads.

sethaj:
Makes the edit section page's monograph dropdown be set to the section's monograph.

Resolves #187

### DIFF
--- a/spec/models/section_spec.rb
+++ b/spec/models/section_spec.rb
@@ -3,9 +3,21 @@ require 'rails_helper'
 describe Section do
   let(:instance) { described_class.new }
   let(:date) { DateTime.now }
+  let(:monograph_id) { '123' }
 
   it "has date_published" do
     instance.date_published = [date]
     expect(instance.date_published).to eq [date]
+  end
+
+  it "has a monograph_id" do
+    instance.monograph_id = monograph_id
+    expect(instance.monograph_id).to eq monograph_id
+  end
+
+  it "indexes the monograph_id" do
+    instance.monograph_id = monograph_id
+    indexed = instance.to_solr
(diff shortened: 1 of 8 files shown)


spec/models/section_spec.rb
```diff
@@ -3,9 +3,15 @@
 describe Section do
   let(:instance) { described_class.new }
   let(:date) { DateTime.now }
+  let(:monograph) { create(:monograph) }
```
val99erie: You don't need to actually create a monograph here.  It will slow down the tests to create objects in fedora.  Instead you can do something like this:
```ruby
let(:monograph_id) { '123' }

it "has a monograph_id" do
  instance.monograph_id = monograph_id
  expect(instance.monograph_id).to eq monograph_id
end
```

spec/indexers/section_indexer_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'rails_helper'
+
+describe SectionIndexer do
+  let(:indexer) { described_class.new(section) }
+  let(:monograph) { create(:monograph) }
```
val99erie: Instead of creating a monograph here, use a hard-coded ID:
`let(:monograph_id) { '123' }`

Issue #152, opened 2016-04-14
Test support for non-English characters in file names


shanitajohn:
**ISSUE** 

Test support for file names in the metadata template with non-English or special characters. 

**ACCEPTANCE CRITERIA**
- [x] When a metatdata template including a file name with non-English or special characters is uploaded, the importer works as intended, and the file opens and is readable within the UI.

**ADDITIONAL INFORMATION**
- [ ] Check import of number 11 in Northwestern metadata sample template
- [x] Check import of A10, file name with non-English characters

val99erie:
In the 'additional information' section, it talks about number 11.  I'm not sure what that requirement is meant to be.  There are no non-English characters in A11.  But I ran the importer with a filename like A10, and it works fine.

PR #211, opened 2016-05-05
File name with non-English characters. Story #152

val99erie:


conorom:
Looks good... I assume some of these characters are from Japanese 101 :)
Merging now.

PR #217, opened 2016-05-11
importer can add sections (and add files to sections)

### LINKED ISSUE #160: Create a way to associate files with a section of the monograph in the importer
**QUESTION**
How is a file associated with a section? 

**STORY**
As an editor, I would like a way to associate a file with a section within the metadata template. The manifest should be validated and the user should be prompted to fix any errors before import. If something was not properly completed in the import process, I should get an error message. 

**ACCEPTANCE CRITERIA**
- [ ] When I properly complete the metadata template and run it through the importer I can see the section I have created within the monograph and the assets I associated with it.
- [ ] If there are any errors, report them to the user.
- [ ] Have the importer report basic metrics on what was imported (Optional)
- [ ] If I run an importer with a known problem I would expect to not have a new record created in Fedora. 
- [ ] If the Relation to Narrative field is blank, the asset should be attached to the monograph directly not to a section.

**NOTES**
The manifest should create the sections.

conorom:
closes #160


lib/import/importer.rb
```diff
@@ -34,18 +54,29 @@ def csv_files
...
+        end
+
         attributes['files'] = attrs['files'].map do |file|
           File.new(find_file(file))
         end
+
         attributes
       end
 
       # TODO: If the file is missing, raise an error?
```
val99erie: You can remove these TODO comments since you implemented the fix

spec/lib/import/csv_parser_sections_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+describe Import::CSVParser do
+  let(:input_file) { File.join(fixture_path, 'csv', 'import_sections', 'tempest_sections.csv') }
+  let(:parser) { described_class.new(input_file) }
+
+  before do
+    # Don't print status messages during specs
+    allow($stdout).to receive(:puts)
+  end
+
+  describe 'initializer' do
```
val99erie: You can delete this test because it is already tested here:
https://github.com/curationexperts/heliotrope/blob/master/spec/lib/import/csv_parser_spec.rb#L13-L17

spec/lib/import/csv_parser_sections_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    it 'has an input file' do
+      expect(parser.file).to eq input_file
+    end
+  end
+
+  describe '#attributes' do
```
val99erie: I think we can just combine this test with the similar test here:
https://github.com/curationexperts/heliotrope/blob/master/spec/lib/import/csv_parser_spec.rb#L22

And then this whole spec file can be deleted.

Issue #195, opened 2016-04-27
Add a data model for sub-brands


mark-dce:
**ACCEPTANCE**
- [x] A press can contain 0 or more sub-brands
- [x] A sub-brand may contain other sub-brands
- [x] A sub-brand and it's children may not contain itself (no loops)
- [x] A sub-brand has a title
- [x] A sub-brand has a description
- [x] A sub-brand can contain 0 or more monographs

val99erie:
I added a follow-up story #224 to improve the UI on the monograph edit form.

I left the following things out-of-scope for this story because they are not needed for the summer beta release:
- Deleting a sub-brand (and updating all the sub-brand's monographs to remove the relationship)
- Faceting by sub-brand on the index page for the press_catalog_controller
- Adding a sub-brand to another sub-brand in the UI (e.g. Add a 'parent' field to the form for a sub-brand?)

PR #223, opened 2016-05-13
Sub-brands. Story #195

val99erie:


val99erie:
Note:  Sub-brands are stored in the relational database, not fedora (similar to presses).

PR #226, opened 2016-05-17
manage users link shown only to certain privledged users

### LINKED ISSUE #121: Remove Users link from menu if user does not have privileges
As an editor/admin of Press 1, I should only be able to see a "Manage Users" link for Press 1. I should not be able to see the "Manage Users" link for any other presses.

Note: Some of these acceptance criteria might already be met, but we should make sure to have specs in place to test all cases.

**ACCEPTANCE CRITERIA**

As an admin of a press...
- [ ] ... when I visit the catalog page for that press, I should see the "Users" link in the menu (the link to manage user roles for this press).
- [ ] ... when I visit the catalog page for a different press (that I'm not an admin for), I should not see the "Users" link.

As an editor of a press...
- [ ] ... when I visit the catalog page for any press (even a press that I'm a member of), I should not see the "Users" link.

As a platform-wide admin...
- [ ] ... when I visit the catalog page for any press, I should see the "Users" link in the menu.

sethaj:
resolves #121

### DIFF
--- a/app/views/shared/_my_actions.html.erb
+++ b/app/views/shared/_my_actions.html.erb
@@ -17,7 +17,12 @@
       <% end %>
 
       <% show_jobs_link = current_user && current_user.platform_admin? %>
-      <% show_users_link = @press && can?(:manage, Role) %>
+      <%
+        show_users_link = @press && (
+          current_user.platform_admin? ||
+          current_user.roles.where(role: 'admin', resource_id: @press.id, resource_type: 'Press').first
+        )
+      %>
       <% show_embargo_link = can? :discover, Hydra::AccessControls::Embargo %>
       <% show_lease_link = can? :discover, Hydra::AccessControls::Lease %>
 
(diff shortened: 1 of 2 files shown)


val99erie:
We have this same problem with several of the admin menu links; The main problem is that in order to correctly determine whether or not a user should have access to the link, we need to know the press.

When the user is on the main catalog page, or the root page, there is no press in scope, but when the user is on a press_catalog page, for example, we do know the press, and we should be able to pass it to the Ability class so that cancan can properly decide the abilities.

In the alexandria code base, we had a similar requirement:   in order to determine certain abilities, we needed to know if a user's IP address corresponded to an 'on-campus' location.

I wonder if we could do something like this for heliotrope?  i.e. Have the base application_controller set the press (if there is a press in scope), and then pass it to the ability class.

Here's the example code from alexandria:
https://github.com/curationexperts/alexandria-v2/blob/master/app/models/ability.rb#L9-L12

and here:
https://github.com/curationexperts/alexandria-v2/blob/master/app/controllers/application_controller.rb#L26-L28

Issue #155, opened 2016-04-14
Store primary creator first and last name separately and concatenated


shanitajohn:
**ISSUE**
Right now heliotrope is concatenating Primary Creator last and first names on ingest - we need to store those separately; concatenate in Solr for searching and for faceting 

**DETAILS**
We probably shouldn't be concatenating first and last name before saving them in Fedora (which means we might need a new model to store complex names).  But, we need to index the concatenated last and first names in Solr (probably separated by a comma), so that we can facet and sort on Primary Creator.

**ACCEPTANCE CRITERIA**
- [x] I should be able to access Primary Creator last name exactly as it was supplied
- [x] I should be able to access Primary Creator first name exactly as it was supplied
- [x] I should be able to facet on Primary Creator in <last name>,+<first name> alphabetical order

val99erie:
Possible predicates:
`::RDF::Vocab::FOAF.family_name`
`::RDF::Vocab::FOAF.givenname`

val99erie:
- [ ] Also, the importer should do the right thing

val99erie:
@shanitajohn When you demo this story, keep in mind that if we have old monographs or assets in the demo environment, the "creator" field for those old records won't show up in the UI.

For new monographs and assets, when you create the records, you should see the first and last name of the creator as separate fields in the edit form, but it will display as a concatenated name in the show view and catalog view.

For old records, the creator field won't show up in the UI at all, unless you go back and edit the record and add a separate first and last name for the creator.

I just wanted to warn you in case someone noticed that the creator fields are missing for the old records.

val99erie:
@shanitajohn & @mark-dce  Since I'm going to be on vacation next week, I won't be able to close this story.  After this pull request gets merged to master, this story can be closed:
https://github.com/curationexperts/heliotrope/pull/235

PR #235, opened 2016-05-20
Separate creator first and last name. Story #155

val99erie:


acozine:
Deployed to sandboxes from the branch, tested, confirmed by @jmcglone, merging.

PR #245, opened 2016-06-17
add a couple of metadata fields and a test for reusing the same asset…

conorom:
… in different sections


sethaj:
This looks good, but it's failing Travis due to something with RuboCop. I'll merge once that's fixed.

PR #247, opened 2016-06-17
added clamav to production

### LINKED ISSUE #246: Add virus checking
Need to add virus checking by adding clamav.
https://github.com/projecthydra/curation_concerns#virus-detection

sethaj:
Resolved #246

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
 language: ruby
 cache: bundler
-sudo: false
+sudo: required
 
+before_install:
+  - sudo apt-get install -qq clamav-daemon libclamav-dev clamav-data
+  - sudo freshclam -v
 before_script:
   - bundle exec bin/setup
 rvm:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,7 +1,10 @@
 language: ruby
 cache: bundler
-sudo: false
+sudo: required
 
+before_install:
```
jmcglone: If clamav install ends up slowing Travis down significantly, we may want to exclude it from this process. I think there is a way to still include it in the Gemfile but when given a `TRAVIS` environment variable, clamav would not get built.  See how Scholarsphere does it: https://github.com/psu-stewardship/scholarsphere/blob/f4e46d132c2042d6cad73115773a179f0f56245e/Gemfile#L19

jmcglone:
@sethaj another question -- do we know happens when a file gets uploaded that is rejected by clamav?

.travis.yml
sethaj: Interesting, I hadn't seen that. Agree that if it slows down tests we should exclude it.

sethaj:
@jmcglone Looks like it happens in Hydra::Works, https://github.com/projecthydra/hydra-works/blob/master/lib/hydra/works/virus_scanner.rb

PR #281, opened 2016-06-29
indexing sort_date for file_sets/assets, #200

### LINKED ISSUE #200: Index normalized dates
**STORY**
As a user, I would like to be able to facet or sort by dates, so I can find assets based on relevant dates.

**DETAILS**
Input fields for dates are not guaranteed to be normalized by the user inputting them (author for metadata templates, editor for UI based input).  So, the system needs to attempt to parse any input dates and normalize them to some standard.  Ruby provides nice support for [ISO-8601 format date-time values](http://ruby-doc.org/stdlib-2.1.0/libdoc/time/rdoc/Time.html#method-i-xmlschema).  So, we probably want to index & store a parsed version of the date that can be used for faceting, searching, and sorting.  It's generally cheapest to do this a single time on ingest or edit of metadata than trying to parse on-the-fly when faceting or sorting.

**ACCEPTANCE**
- [ ] The original user-supplied date is retained as an unmodified string
- [ ] The system saves a parsed version of the date (when possible) in ISO-8601 format
- [ ] Unparseable date strings do not get indexed into the ISO-8601 date index (or get index as nil)

sethaj:
Resolves #200 

This enforces a sort_date format of YYYY-MM-DD, so it won't accept YYYY or YYYY-MM. Is that ok with everyone?

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -12,9 +12,11 @@
 //
 //= require jquery
 //= require jquery_ujs
+//= require jquery-ui/datepicker
 //= require turbolinks//
 // Required by Blacklight
 //= require blacklight/blacklight
 
 //= require 'edit_users'
+//= require 'file_set_sort_date'
 //= require_tree .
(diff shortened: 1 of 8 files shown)


jmcglone:
Because this is just for sorting and not display (right?), I think forcing the YYYY-MM-DD is fine.

PR #376, opened 2016-08-10
Custom navbar to match Elderberry wireframes

### LINKED ISSUE #315: Customize header and navigation
Modify the CC header and navigation menu so that it matches the nav menu and header wireframed in Elderberry.

jmcglone:
Closes #315 

In addition to modifying partials to bring the header, navigation, and press header inline with Elderberry wireframes, this PR:

- Introduces a class based approach to handle "press themes" in Fulcrum. 
- Search is removed from the top-level now and only available at the Press level and below.
- Maintains the parent press navigation at the Monograph and Asset level

### DIFF
--- a/app/assets/javascripts/short_monograph_description.js
+++ b/app/assets/javascripts/short_monograph_description.js
@@ -8,7 +8,7 @@ jQuery(function(){
 
         $(this).html(
             t.slice(0,500)+'<span>... </span><a href="#" class="more">More</a>'+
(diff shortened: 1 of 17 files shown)


jmcglone:
screen shot 2016-08-10 at 3 29 50 pm

screen shot 2016-08-10 at 3 31 48 pm

screen shot 2016-08-10 at 3 33 21 pm

screen shot 2016-08-10 at 3 33 26 pm

screen shot 2016-08-10 at 3 34 40 pm

PR #382, opened 2016-08-11
change resque_namespace per environment

### LINKED ISSUE #380: Set prefix/namespace for redis keys per environment
See comment :)

conorom:
closes #380

### DIFF
--- a/config/initializers/resque_config.rb
+++ b/config/initializers/resque_config.rb
@@ -3,4 +3,4 @@ config = YAML.load(ERB.new(IO.read(File.join(Rails.root, 'config', 'redis.yml'))
 Resque.redis = Redis.new(host: config[:host], port: config[:port], thread_safe: true)
 
 Resque.inline = Rails.env.test?
-Resque.redis.namespace = "#{Settings.resque_namespace}:#{Rails.env}"
+Resque.redis.namespace = Settings.resque_namespace || "heliotrope"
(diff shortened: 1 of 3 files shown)


config/initializers/resque_config.rb
```diff
@@ -3,4 +3,4 @@
 Resque.redis = Redis.new(host: config[:host], port: config[:port], thread_safe: true)
 
 Resque.inline = Rails.env.test?
-Resque.redis.namespace = "heliotrope:#{Rails.env}"
+Resque.redis.namespace = "#{Settings.resque_namespace}:#{Rails.env}"
```
grosscol: Rails.env shouldn't be required here.  It will pull from whatever you put in the config/settings/production.yml file, and that file can be tailored per deployment.  I would also add a default with ||
```
Resque.redis.namespace = Settings.resque_namespace || "heliotrope:defaulty"
```

but you might want to open up pry somewhere and see what values that aren't specified in settings ends up being.  e.g. Settings.foo.   I'm assuming it's nil or something falsy, but I may be unhappily surprised here.

conorom: Good point. Especially once I differentiated dev and test which were both just "heliotrope" previously.

grosscol:
Looks good to me.  Squash the commits since the second one is a fixup?

PR #383, opened 2016-08-11
add custom footer at the press, monograph, and asset level

### LINKED ISSUE #314: Customize footer
Modify the CC footer so that it matches the footer wireframed in Elderberry.

jmcglone:
Resolves #314 

NOTE: adds a new column `press_url` to the `presses` table to capture the presses url (ie http://www.press.umich.edu).  So you will need to run:

`bundle exec rake db:migrate`

then

`bundle exec rake db:seeds`

to populate the updated presses table.

TODO: @sethaj I can't figure out how to get the press logo (`logo_path`) and the press url (`press_url`) into the presenters for the monograph and asset, or what the best way to go about doing this is (probably don't need to be indexed in solr, do they?). The footer design looks for the press logo and url and inserts them in the footer if they exist.

### DIFF
--- a/app/assets/stylesheets/heliotrope.scss
+++ b/app/assets/stylesheets/heliotrope.scss
@@ -5,6 +5,54 @@
   margin-bottom: 0;
 }
 
+// footer
+// generic press
+footer.press {
+  background-color: #262626;
+  color: #fff;
+  margin-top: 5em;
+
+  img {
+    width: 100px;
+    height: auto;
+  }
+
+  .col-wrap {
+    overflow: hidden;
+  }
+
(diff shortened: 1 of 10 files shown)


sethaj:
In the presenter you already have

```
def subdomain
   Array(solr_document['press_tesim']).first
end
```
So you should be able to do something like

```
def press_logo
  Press.find_by(subdomain: subdomain).logo_path
end
```

Want to try it and role it into this PR, or make a new story?

jmcglone:
@sethaj thanks! I'll give that a shot and roll it into this PR if it works.

PR #403, opened 2016-08-18
333: Fixed page title for asset page.

### LINKED ISSUE #333: Page title for asset pages should be the asset title
Change the page title to use the asset title.

Currently, the page title uses the file name. The file name should never be public-facing.

screen shot 2016-08-01 at 7 34 02 pm

The page title should instead be: Photograph of Nina in The Seagull

gkostin1966:
Resolves #333


gkostin1966:
Jumped the gun, forgot to write a test!

Issue #400, opened 2016-08-16
IIIF and imagemagick


sethaj:
It's possible that we're seeing performance issues/memory problems with large tiffs and the riiif gem which uses imagemagick to create IIIF tiles in staging.

- [x] Examine and stress test riiif with large images on the staging server
- [x] if needed, come up with an alternate way  to extract images that doesn't use imagemagick, or decide on a different IIIF server solution that doesn't use the riiif gem.
- [x] Stress test the new solution in staging to see if we're still having memory issues

sethaj:
Maybe also don't hit fedora for these if possible and use solr instead?

https://github.com/mlibrary/heliotrope/blob/master/config/initializers/riiif_initializer.rb#L13

sethaj:
Make sure we're caching IIIF tiles if we can.

sethaj:
With the change in place, here's 6 concurrent users hitting the image-service route requesting random tiles for 60 seconds (urls generated with the sample_iiif_urls rake task #425). There's no real impact on memory anymore.

```
~   ᐅ siege -f ~/urls.txt -i -c 6 -b -t60S
** SIEGE 3.0.8
** Preparing 6 concurrent users for battle.
The server is now under siege...
Lifting the server siege...      done.

Transactions:		         474 hits
Availability:		      100.00 %
Elapsed time:		       59.25 secs
Data transferred:	      271.42 MB
Response time:		        0.74 secs
Transaction rate:	        8.00 trans/sec
Throughput:		        4.58 MB/sec
Concurrency:		        5.94
Successful transactions:         474
Failed transactions:	           0
Longest transaction:	        8.29
Shortest transaction:	        0.02
```

PR #425, opened 2016-08-24
rake task to generate urls to stress test riiif and netpbm

sethaj:
This relates to #400 (testing riiif to make sure there are no memory problems)


jmcglone:
@sethaj if you think it would be useful to others (not sure tho) don't forget the info in the comments could also be added to the wiki!

Issue #389, opened 2016-08-13
Implement streaming video


jgmorse:
Use (or help complete) this gem: https://github.com/UCSCLibrary/hydra_hls

As discussed here:
https://groups.google.com/forum/#!msg/hydra-community/sSIzviZQiPI/PPZd-mgEAQAJ

mbakeryo:
A&E will tell us when we need to do this (due to # of simultaneous viewing streams). Even if we do the Deaf Studies journal (which entirely videos), we still won't be close to that scale.

PR #458, opened 2016-09-06
temporary byte stream fix for derivatives, #401, until x-sendfile is …

### LINKED ISSUE #401: Investigate video playback on iPad and Kindle Fire
Current behavior: When I view a video asset page on an iPad, Kindle Fire, iPhone, or Safari, the video asset does not playback.

Expected behavior: When I view a video asset page on an iPad or Kindle Fire, the video asset should have playback controls and play when I touch the play button.

- [ ] Video plays back on iPad, Kindle, and iOS (when content-type is set correctly)
- [ ] Tested across other touch screen and webkit devices in browserstack and works (when content-type is set correctly)

sethaj:
…installed or #389 happens

Temporarily fixes #401 for development. I don't think it's a  good idea to rely on this for production:

https://github.com/mlibrary/heliotrope/commit/c953e533040702a5177e0a36d6413f058d85e413#diff-5b1717b5c25ba511288b5bb06c831667R81

### DIFF
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -1,9 +1,18 @@
 class DownloadsController < ApplicationController
   include CurationConcerns::DownloadBehavior
 
+  attr_accessor :file_mime_type
+
   def show
     if thumbnail? || video? || sound? || allow_download?
-      super
+      # override some CC DownloadBehavior
+      case file
+      when String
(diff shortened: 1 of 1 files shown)


sethaj:
We're going to go with x-sendfile instead of doing this on staging, preview and production. This means iOS and Safari won't work in dev and test (since there's no apache in front), but should work otherwise.

PR #464, opened 2016-09-12
add X-Sendfile header for videos, #401

### LINKED ISSUE #401: Investigate video playback on iPad and Kindle Fire

sethaj:
Resolves #401 

This needs mod_xsendfile installed on Apache as well as some Apache config to work. What this means is iOS and Safari will be able to view video in production, staging, and preview, anywhere that has Apache in front of the app. This will not work for dev and test. iOS and Safari will not be able to view HTML5 video in dev and test.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -10,3 +10,4 @@ Metrics/CyclomaticComplexity:
 Metrics/PerceivedComplexity:
   Exclude:
     - 'app/models/riiif/file.rb'
+    - 'app/controllers/downloads_controller.rb'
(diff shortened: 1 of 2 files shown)


jmcglone:
@sethaj sorry, I'm confused by our own terminology here...

what's the difference between dev and test when compared to staging and preview?

sethaj:
Dev is your development environment, so local development. Test is where we run tests, so rspec and travis. This is all Rails terminology. Pretty much anything other than local development or rspec/travis is "production" as far as Rails in concerned. Of course for us that's not really true, staging is staging and preview is preview, but internally Rails considers them "production".

PR #576, opened 2016-11-03
Resolves #489: Ability to navigate next and previous from asset page.

### LINKED ISSUE #489: Add ability to navigate to the next and previous asset from the asset page
As a general public user, while on an asset page, I should be able to navigate to the next asset, or move to the previous asset, in the relevant results list, so I don't have to go back to the main results page.

Is this straight-forward or is additional work needed?

For example, do we need to work out what assets this navigation will be applied to? Will facets, sort order, etc., be considered in this navigation.

gkostin1966:
You can now navigate the monograph's assets (file_sets) from the asset page using previous and next navigation buttons in the order they appear in the monograph.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -82,6 +82,7 @@ group :development, :test do
   #gem 'fcrepo_wrapper', '0.6'
   #gem 'solr_wrapper', '~> 0.16'
   gem 'rspec-rails'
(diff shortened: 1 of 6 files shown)


jmcglone:
@gkostin1966 can you upload some screenshots?

PR #585, opened 2016-11-07
Resolves #584: Remove More/Less from Asset Description.

### LINKED ISSUE #584: Remove More/Less from Asset Descriptions
A user should be able to see all of the asset descriptive copy on the asset page without having to click a More/Less link.

Verify that the breaks in the descriptive copy are restored.

gkostin1966:
Created .short_description scss class for more/less behavior.

### DIFF
--- a/app/assets/javascripts/short_monograph_description.js
+++ b/app/assets/javascripts/short_monograph_description.js
@@ -1,6 +1,6 @@
 jQuery($(document).on('turbolinks:load', function (){
 
-    var minimized_elements = $('.description');
+    var minimized_elements = $('.monograph-metadata .description');
 
     minimized_elements.each(function(){
         // TODO: Threshold should probably be passed in as a parameter.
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/short_monograph_description.js
```diff
@@ -1,6 +1,6 @@
 jQuery($(document).on('turbolinks:load', function (){
 
-    var minimized_elements = $('.description');
+    var minimized_elements = $('.short_description');
```
jmcglone: @gkostin rather than change the class on the div, you should be making your jquery selector more specific -- the monograph `.description` we want to shorten only ever appears in the parent div `.monograph-metadata` -- so if you change this line to ` minimized_elements = $('.monograph-metadata .description');` and change the `index.html.erb` file back to its original `description` class, it should work.

gkostin1966: Sounds good to me.

PR #640, opened 2016-11-22
Resolves #599: Updating the file of an existing file_set should work …

### LINKED ISSUE #599: Leaflet caching breaks when replacing a file
When a FileSet's file is changed, the Leaflet cache should be cleared.

- [ ] The Leaflet viewer shows the new file as soon as the relevant Resque jobs are finished

sethaj:
…in riiif and leaflet

1) Override Riif::Image.cache_key to include the file_set's timestamp when caching image tiles.
If the file/image changes, then the old cache_keys will no longer work and new ones will be
generated. The cache is cleaned periodically via a rake task and cron to clean up these old
cache tiles.

2) When a file is characterized we need to delete riiif's cache of it in tmp/network_files. If it's a new file it won't have one and won't matter.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -89,4 +89,5 @@ group :development, :test do
   gem 'sqlite3'
   gem 'factory_girl_rails'
   gem 'capybara'
+  gem "fakefs", require: "fakefs/safe"
 end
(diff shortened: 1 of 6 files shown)


gkostin1966:
Is it possible to write a test to verify the deletion of the file?

Also, what happens if file_set.characterization_proxy.save! throws?  Does the frame work already handle these exceptions?

Otherwise it looks great, ship it!

sethaj:
file_set.characterization_proxy.save! happens in CC. But I added a test to verify the (cached) file deletion.

PR #659, opened 2016-12-08
persist fedora data in development

sethaj:
I've tested this many different ways in my dev environment and it seems to work.


sethaj:
So this doesn't work with vagrant. So instead of having the fedora dir in the .fcrepo_wrapper config file, it can just be passed on the commend line like: `bundle exec fcrepo_wrapper -p 8984 --no-jms --fcrepo_home tmp/fcrepo` which works the same.

PR #667, opened 2016-12-16
Resolves #368

### LINKED ISSUE #368: Facet on Content Type (subfacet of Format)
Is there a way for when faceting on Format, a user could then be presented with subcategories (Content Type) to choose from?

Research how this would work and what it would look like.

gkostin1966:


sethaj:
I think this all looks fine from a UI standpoint except for one little issue.
facet-pivot

The "150" count for images is slightly off center. In the code we have:

`<span class="facet-count" style="width: 24px;">150</span>`

but for the others (video, audio) we have: 

`<span class="facet-count">3</span>`

Getting rid of that `width: 24` makes it looks fine. I'm trying to find where that style gets generated, but no luck so far. If we can get rid of that I think this PR is fine.

PR #677, opened 2017-01-04
Hide download button to buy time for a real solution.

gkostin1966:


app/assets/stylesheets/heliotrope.scss
```diff
@@ -300,6 +300,23 @@ footer.press {
...
+  ::-webkit-media-controls-panel {
+    width: calc(100% + 30px); /* Adjust as needed */
+  }
+  // by hiding the overflow of the enclosure
+  ::-webkit-media-controls-enclosure {
+    overflow:hidden;
+  }
+  // but this won't work for non-shadow-DOM browsers
+  // so this is a temporary fix to buy time.
+
```
jmcglone: @gkostin1966 because this also impacts Safari, you could do a media query for Safari 7+ (that will also catch Chrome 22-28, but that doesn't matter b/c there isn't a DL button in Chrome 28). Try this at line 319:

```
@media \\0 screen {
  ::-webkit-media-controls-panel {
    width: 100% !important;
  }
}
```

PR #698, opened 2017-01-17
Resolves #693: Blacklight Helper Rspecs

### LINKED ISSUE #693: Blacklight Helper Rspecs

gkostin1966:
Thanks Valerie :)

### DIFF
--- a/app/helpers/facet_helper.rb
+++ b/app/helpers/facet_helper.rb
@@ -52,6 +52,9 @@ module FacetHelper
     render_markdown_as_text(value)
   end
 
+  ############################################################################
+  ## Begin Blacklight Helper Override...
+
   ##
   # Determine whether a facet should be rendered as collapsed or not.
   #   - if the facet is 'active', don't collapse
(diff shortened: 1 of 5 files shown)


app/helpers/facet_helper.rb
```diff
@@ -60,27 +63,89 @@ def markdown_as_text_facet(value)
   #
   # @param [Blacklight::Configuration::FacetField] facet_field
   # @return [Boolean]
+  # def should_collapse_facet? facet_field
+  #   !facet_field_in_params?(facet_field.field) && facet_field.collapse
+  # end
+  #
```
val99erie: In this file, I don't think you should include the original methods in the comments.  Including the old methods make the file more cluttered and that info may go out-of-date quickly if you upgrade blacklight.

I think it's enough to just say that you are overriding methods from blacklight (and why you need to override), and then future devs will know that they need to look in blacklight to find the original methods.

gkostin1966: Sounds good to me.  I'll trim down the comments.

Issue #673, opened 2017-01-03
Refactor slow tests


sethaj:
It might be time to explore ways to refactor some of our slowest tests. Use `rspec --profile` to get this output.

```

Top 10 slowest examples (122.03 seconds, 53.6% of total time):
  Import::Importer#run when the importer runs (reversing the rows) successfully imports the sections (unreversed) and files (reversed)
    21.26 seconds ./spec/lib/import/importer_spec.rb:88
  Monograph Catalog Facets sections shows sections in intended order
    20.86 seconds ./spec/features/monograph_catalog_facets_spec.rb:32
  Import::Importer#run when the importer runs successfully imports the monograph, sections and files
    19.92 seconds ./spec/lib/import/importer_spec.rb:46
  FileSet Browse Navigating forward/backward navigation arrows
    16.2 seconds ./spec/features/show_file_set_spec.rb:34
  Monograph catalog sort FileSet results set displays 20 results by default and is ordered as expected
    15.91 seconds ./spec/features/monograph_catalog_sort_spec.rb:27
  Monograph Catalog Search searches the monograph catalog page
    6.34 seconds ./spec/features/monograph_catalog_search_spec.rb:17
  Create a file set as a logged in user should have css "ul.nav.nav-tabs li a"
    5.81 seconds ./spec/features/create_file_set_spec.rb:13
  Monograph Catalog Facets all facets shows the correct facets
    5.47 seconds ./spec/features/monograph_catalog_facets_spec.rb:127
  Monograph Catalog Facets sections shows italics (emphasis) in section facet links
    5.3 seconds ./spec/features/monograph_catalog_facets_spec.rb:93
  CurationConcerns::Actors::AssetActor has the user's email address by default
    4.97 seconds ./spec/actors/curation_concerns/asset_actor_spec.rb:13

Top 10 slowest example groups:
  FileSet Browse
    16.2 seconds average (16.2 seconds / 1 example) ./spec/features/show_file_set_spec.rb:3
  Monograph catalog sort
    15.91 seconds average (15.91 seconds / 1 example) ./spec/features/monograph_catalog_sort_spec.rb:3
  Monograph Catalog Facets
    8.39 seconds average (33.55 seconds / 4 examples) ./spec/features/monograph_catalog_facets_spec.rb:3
  Create a file set
    5.81 seconds average (5.81 seconds / 1 example) ./spec/features/create_file_set_spec.rb:3
  Import::Importer
[...]

PR #699, opened 2017-01-17
Refactored a spec to speed it up. Story #673.

val99erie:
I rewrote the spec to create fewer FileSet records in fedora.
Before changes, the spec took 25-27 seconds on my laptop.
After changes, it takes 8-10 seconds.


val99erie:
Ping @conorom for code review.  Could you take a look at this change and make sure I'm not violating the original spirit of the spec?  The sort orders are still the same as before, but there are fewer records per page.

PR #742, opened 2017-01-30
Updating to CC 1.7.0

### LINKED ISSUE #675: Remove minter_state.rb
We have a file app/models/minter_state.rb that existed because of a bug in the way CC included ActiveFedora::Noid. Looks like this was fixed (in October!) so we should be able to remove this file now provided we're using a new enough version of CC.

See https://github.com/mlibrary/heliotrope/issues/292
See https://github.com/projecthydra/curation_concerns/pull/1061

### LINKED ISSUE #676: Update to CC 1.7.0
Or maybe wait until 2.0 (rc1 is out now).

sethaj:
Resolves #676 
Resolves #675 

This CC update includes workflows with sipity. To install:

```
# Run the migrations (adds sipity/workflow tables)
bundle exec rake db:migrate

# Load the default workflow
bundle exec rake curation_concerns:workflow:load

# Add sipity_entity rows for each Work that's already been imported
bundle exec rake heliotrope:add_sipity_entities
```


sethaj:
PR #764 should probably be merged and the changes rolled into this branch before closing this PR. It seem like doing it the other way could be harder/more confusing.

PR #752, opened 2017-02-02
When we cache GA data we want all of it, not only 1 month and not onl…

### LINKED ISSUE #748: Check inconsistent pageviews in GA
Data displayed in monograph and asset pages does not match what is displayed in google analytics account. Users are also reporting seeing pageview stats going up and down.

sethaj:
…y 1000 rows

Resolves #748


jmcglone:
@sethaj so we were only getting 1-month at a time?

sethaj:
@jmcglone Looks that way... and then also only getting 1000 records max. I couldn't find these defaults set anywhere, but the GA API documentation is kind of... comprehensive. Maybe I just missed it.

PR #794, opened 2017-03-08
fix facet bugs #772

### LINKED ISSUE #772: Fix multi-word entries from breaking facet link
When a user clicks on a multi-word facet link (e.g. on screen talent), it should not bring up a no results found page. 

For an examples of this behavior, see:
1. Indiana's The Great Salish Sea music video and click on the primary role of on screen talent.
2. Indiana's Just Want to be With You and click on the keyword Walkin' Jim Stoltz

conorom:
Resolves #772


sethaj:
We'll need to remember to reindex_everything after this is merged.

PR #835, opened 2017-03-24
only show 'Exclusive to Fulcrum' in exclusivity facet

### LINKED ISSUE #808: Change Exclusivity facet to include Exclusive to Fulcrum
- [ ] Change  "Does not appear in book" to "Exclusive to Fulcrum"
- [ ] Remove "Appears in book"

conorom:
Resolves #808


sethaj:
I think something might be off with the importer on this branch.

Here's what we have now. A fresh import of tiny_michigan from the master branch:

imported-on-master

When you switch to this branch ( 808_change_exclusivity_facet) so far so good:

inported-on-master-switched-to-branch

But then I tried to do a fresh import of tiny_michigan (after deleting the old monograph) on the 808_change_exclusivity_facet branch, the facet is gone:

new-import-on-branch

I've tried to be careful about restarting resque when I was doing the imports on different branches, but I might have messed something up. Can someone confirm imports work correctly on this branch?

conorom:
Apologies, I should have pointed out the metadata template is changing so this field is simply called "Exclusive to Fulcrum" and not the massive long BP/P thing. I was going to make it work with both but figured it was unnecessary mess.

PR #855, opened 2017-03-31
renderer option to preseve newlines only; fix spec

### LINKED ISSUE #814: Unwanted Markdown in WebVTT
'Greater than' signs in the WEBVTT are triggering markdown 'blockquotes' formatting (looks like email reply text).
https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet

IIRC I rendered the transcript as markdown to allow newlines to persist, which is essential, obviously. I didn't notice any markdown issues at that time.

This is the first thing to try:
http://stackoverflow.com/questions/28397943/preserve-newline-in-text-area-with-ruby-on-rails

screen shot 2017-03-20 at 1 20 23 pm

conorom:
Resolves #814


gkostin1966:
I'm leaning toward creating a transcript_attribute_renderer, but maybe that is overkill.

PR #857, opened 2017-04-04
Resolves #845: Platform Admin Edit Monograph

### LINKED ISSUE #845: Platform Admin - edit a monograph
- [x] Login as platform admin
- [x] Choose publisher
- [ ] See list of monographs with status (draft or published)
- [x] Select monograph to edit
- [x] Save (keeps whatever state it is in - draft or published)

gkostin1966:


gkostin1966:
Created story card #858 to address monograph draft or published status and set it to blocked until after port to Hyrax.

Currently all monographs are published.

jmcglone:
Overall this is a great start. I am unclear about the sidebar navigation items, which diverge from what we have agreed upon when white boarding and wire framing the dashboard (repository, profile, users, dashboard are all labels we have not used before). As a refresher, we want something like this for the dashboard navigation, with "Publishers" added after "Home" and before "Books":

screen shot 2017-04-05 at 9 31 35 am

So I would:

- Change "Dashboard" to "Home". This would also fix the weirdness of having dashboard/dashboard in the URL path.
- Change "Repository" to "Reports". It could be the beginning of some kind of reporting tool for admins.
- Get rid of "Profile" for now.
- I think it's okay to keep Users for now.
- Add a "Books" section (empty for now?) which starts to fulfill the request in the card.

Other thoughts:

- We need to definitely improve the user flow for getting to edit a monograph. Right now it is four clicks from the dashboard Publisher page and you would need to know exactly where to click on a page to get there. Maybe that's a new card though with specific details.

PR #920, opened 2017-04-27
Fixes #869, ActiveFedora::SolrService deprecation warnings

### LINKED ISSUE #869: ActiveFedora::SolrService deprecation warnings
We should clean up all of these deprecation warnings by adding a :rows parameter whenever ActiveFedora::SolrService is called. Which might be in quite a few places... Here's an example from the log:

```
W, [2017-04-11T13:33:11.856058 #22820]  WARN -- : Calling ActiveFedora::SolrService.get without passing an explicit value for ':rows' is not recommended. You will end up with Solr's default (usually set to 10)
Called by /hydra-dev/heliotrope-staging/releases/20170411163212/app/controllers/downloads_controller.rb:30:in `file_set_doc'
```

sethaj:
I think I got them all...

### DIFF
--- a/app/indexers/monograph_indexer.rb
+++ b/app/indexers/monograph_indexer.rb
@@ -13,7 +13,7 @@ class MonographIndexer < CurationConcerns::WorkIndexer
   end
 
   def existing_filesets
-    existing_monograph_doc = ActiveFedora::SolrService.query("{!terms f=id}#{object.id}")
+    existing_monograph_doc = ActiveFedora::SolrService.query("{!terms f=id}#{object.id}", rows: 1)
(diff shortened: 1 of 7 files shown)


app/indexers/monograph_indexer.rb
```diff
@@ -13,7 +13,7 @@ def generate_solr_document
   end
 
   def existing_filesets
-    existing_monograph_doc = ActiveFedora::SolrService.query("{!terms f=id}#{object.id}")
+    existing_monograph_doc = ActiveFedora::SolrService.query("{!terms f=id}#{object.id}", rows: 10_000)
```
conorom: only expecting one monograph result here?

sethaj: Oops, yes. Nice catch, thanks.

PR #933, opened 2017-05-05
Update cozy sun bear

### LINKED ISSUE #926: Update heliotrope's version of cozy-sun-bear
- [ ] We should be using the HEAD of the master branch of CSB
- [ ] modify show page to correctly display so `div.#reader` does not have inline height and width

jmcglone:
Resolves #926 

We may want to refactor how fonts are incorporated into the asset pipeline. However, this may be fixed by refactoring how fonts are handled in CSB itself.


jmcglone:
Closing pull request to test @respinos gemification of CSB

PR #934, opened 2017-05-08
Resolves #912: Publishers Summary Page Changes

### LINKED ISSUE #912: Publishers Summary page cleanup
- [x] Remove "Search" option from each publisher row 
- [x] Include "Edit" button for each publisher row - this should take you to the form created from #913 with fields filled out with existing data for that publisher.
- [x] Make each column heading sortable. Default by Publisher Name in alpha order.

gkostin1966:


jmcglone:
@gkostin1966 Screenshot?

app/views/dashboard/_publishers.html.erb
```diff
@@ -23,3 +23,75 @@
     <% end %>
   <% end %>
 </table>
+<script>
```
jmcglone: @gkostin1966 @sethaj @conorom General housekeeping question - are we okay with inline JS like this, or should this be stored and minified with the rest of the app's JS in the `assets` directory? Maybe the file would be called something like  `/app/assets/javascripts/dashboard_sort.js`?

PR #1018, opened 2017-06-12
Resolves #1014 Edit FileSet Page Video

### LINKED ISSUE #1014: Edit FileSet Page doesn't load for Videos
Looks like Hyrax loads a view partial on the edit page (thumbnail for images anyway).
For videos this prevents the page from loading as it looks for transcript data etc.

- [ ] no error 500 on edit page for video FileSets under Hyrax

gkostin1966:
partial media_display's local variable for presenter is 'file_set'


sethaj:
Could this PR be expaned to make sure all the media types are updated if needed or should we open new issues?

https://github.com/mlibrary/heliotrope/blob/c3a7bb66ae39a31664fc592cb9e683e14222bb04/app/views/hyrax/file_sets/media_display/_audio.html.erb#L4

gkostin1966:
Yeah, we can expand it to all media types.

Issue #970, opened 2017-05-25
Investigate FileSetsController/Jobs/Hyrax


sethaj:
We have a pretty ugly shim in the FileSetController that we need to get rid of:

https://github.com/mlibrary/heliotrope/blob/master/app/controllers/curation_concerns/file_sets_controller.rb#L98

It seemed to be needed in CC, but probably (?) not in Hyrax. Once we have a stable Hyrax we should revist this.

sethaj:
The PR for this is 1026

PR #1026, opened 2017-06-14
hyrax changes how file_sets are added to monographs

sethaj:
We used to be able to simultaniously upload a file for a file_set and add it's metadata.  With the "hyrax way" this is no longer the case. Now you first upload the file from the work (monograph) edit/create page, then later add metadata to the uploaded file_set from the file_set edit page.

This PR gets us more "hyrax like" with that workflow. That's probably right, but maybe we want to diverge from hyrax with this feature(?)


sethaj:
Resolves #970

Issue #1117, opened 2017-08-08
Remove scrolling attribute from embed codes


conorom:
`scrolling='no'` is not HTML5 compliant, use `overflow:hidden` instead
https://stackoverflow.com/questions/5048793/the-scrolling-attribute-on-the-iframe-element-is-obsolete-use-css-instead

- [ ] test in epub viewer
- [ ] add to embed code in file_set_presenter.rb
- [ ] change spec

PR #1133, opened 2017-08-15
tweak embed style/codes; reorder bowser; Able Player bugfix

conorom:


conorom:
Uh, this resolved #1117.
Bit late now. Just want to see what happens in Waffle.

Issue #821, opened 2017-03-23
Cover image should use its own alt text


jmcglone:
A monograph's representative media (ie, cover image) should be given alt text and the input for this metadata field should be used in the `alt` attribute of the `img` tag wherever the representative media is rendered in the app. Currently the cover image is rendered on the Press catalog page and Monograph catalog page. It is also rendered on the Fulcrum home page, however this page does not currently connect to Solr because it is a static page.

Currently the cover image is using the title of the Monograph as alt text.

- [x] if alt text is entered for the representative media, use that field in the `alt` attribute of the `img` tag when rendering the object on the monograph catalog page and the press catalog page.
- [x] if alt text file is empty for the representative media, use the title of the monograph in the `alt` attribute of the `img` tag on the monograph and press catalog pages.
- [x] add alt text for each cover image currently in production (Michigan, Minnesota, Northwestern, Indiana)

This may also be necessary in order to get this working on the Press catalog page:

- ~~[ ] ? modify press catalog page so that it calls the representative media asset instead of a hardcoded link to the cover image via IIIF ?~~ (Doesn't look like this is necessary)

moseshll:
@jmcglone I believe I have the code changes near completion, but as a clarification for checkbox 3 ("add alt text…), would this be something done via reingest, in code, or manually?

jmcglone:
@moseshll It would be manually entered by editing the representative media asset's metadata (via the UI) and simply adding the alt text to the "Alt text" field.

PR #1174, opened 2017-08-30
Addresses #821 cover images alt text.

moseshll:
Monograph catalog page only tested via a hack because my dev machine is using placeholders instead of images for cover pages.


jmcglone:
@sethaj looks good to me. Ok to merge?

Issue #1162, opened 2017-08-25
Add search results highlighting


gkostin1966:
As a reader, when I conduct a search inside an EPUB and I follow a link in the results list, I want the matched terms to be highlighted on the page so I can easily find my matched terms.

- [ ] Only matched terms/phrases are highlighted in the EPUB
- [ ] The highlighted terms/phrases match the searched term/phrase exactly

jmcglone:
![screen shot 2017-08-29 at 8 02 03 am](https://user-images.githubusercontent.com/1686111/29820140-6ad315b4-8c91-11e7-8c10-a9254d0d18c7.png)

PR #1185, opened 2017-09-06
Work towards resolving #1162, highlighting search terms in CSB. One of

sethaj:
the requirements for doing that is CFI ranges which are added here.
This also fixes "more than on hit per DOM element" where before only
the first search term was returned for a DOM element, now all the search
terms/hits are returned.

### DIFF
--- a/app/services/e_pubs_search_service.rb
+++ b/app/services/e_pubs_search_service.rb
@@ -28,12 +28,25 @@ class EPubsSearchService
     end
   end
 
-  def snippet(e, q)
-    p0 = e.text.downcase.index(q.downcase)
-    p1 = p0 + q.length
+  def snippet(e, p0, p1)
     before = e.text[p0 - 30..p0 - 1]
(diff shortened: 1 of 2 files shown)


app/services/e_pubs_search_service.rb
```diff
@@ -28,14 +28,27 @@ def walk_up_dom(el, cfi, i)
     end
   end
```
gkostin1966: I spent too much time trying figuring out what all the letter variables stand for:

el = element
cfi = 
i = index
e = element
p0 = point0 a.k.a. lower bound a.k.a. left boundary
p1 = point1 a.k.a. upper bound a.k.a. right boundary
30 = offset before query and offset after query a.k.a. result window size a.k.a. PADDING = 30
q = query
stm = select from match
rs = resultset
els = elements

also, not sure exactly how find_hits works, but I know it is recursive,  looks like tails recursion, should I be concerned about stack overflow?

-Greg

Issue #959, opened 2017-05-18
FileSet file size not set


conorom:
... is often missing. It always seems to be missing if you swap the file out with the `<FileSet URL>/edit` form, but it's also missing from some FileSets that should have been imported the normal way and never touched since. We don't set this in our `file_set_indexer.rb` so we're relying on underlying gems to do it. "Maybe Hyrax will fix it", he says in the manner to which he has become accustomed. We need a label for that.

- [ ] investigate this
- [ ] fix this

sethaj:
PR #1299 might help with this.

sethaj:
This looks to be what always happens with multi-valued fields in fedora: they are not ordered and so you never know which one you're going to get.

Staring with a FIleSet and then updating with different "versions":
```
[ after uploading another file or "version" via the GUI ] 
[3] pry(main)> f.original_file.file_size
=> ["687198", "1827257"]
[ after uploading yet another file or "version" via the GUI ] 
[5] pry(main)> f.reload.original_file.file_size
=> ["1545542", "687198", "1827257"]
[6] pry(main)> [8] pry(main)> f.to_solr["file_size_lts"]
=> "1545542"
[ this is different than the value actually in solr's GUI index, which is 1827257 ]
[ after uploading yet another file or "version" via the GUI ] 
[14] pry(main)> f.reload.original_file.file_size
=> ["1545542", "687198", "1827257", "2026097"]
[15] pry(main)> f.to_solr["file_size_lts"]
=> "1545542"
```

conorom:
I do think this particular issue was from our presenter grabbing the file_size from the old 32-bit int value, when newly added items were having their file_size indexed with the new long value. So it looked like file_size wasn't set sometimes, but it was actually always set in Fedora. I mean, this ticket wasn't about weird changing sizes or ordering.

sethaj:
Yes, true. I got a little side tracked with the "Edit" part of it. I really think PR #1299 and a reindex in production should fix this though. I'll test it out.

sethaj:
What's the current file_size_lts of FileSets in solr?

```
Monograph.all.to_a.each do |m|
  p m.title
  m.ordered_members.to_a.each do |f|
    if f.original_file.present?
      doc = ActiveFedora::SolrService.query("{!terms f=id}#{f.id}", rows: 1).first
      if doc["file_size_lts"].blank?
       puts "#{f.id}"
      else
       puts "\t\t#{doc['file_size_lts']}"
      end
    end
[...]

PR #1299, opened 2017-10-23
post-Hyrax Solr file_size field

conorom:
Will require reindex for content imported pre-Hyrax

### DIFF
--- a/app/presenters/hyrax/file_set_presenter.rb
+++ b/app/presenters/hyrax/file_set_presenter.rb
@@ -217,7 +217,7 @@ module Hyrax
     end
 
     def file_size
-      solr_document['file_size_is']
+      solr_document['file_size_lts']
     end
 
     def last_modified


sethaj:
This is related to #959 ?

conorom:
Could well be. Very likely. When I said "maybe Hyrax will fix it" it's because one of the original issues that caught my eye in CC was tagged "Move to Hyrax", which I assumed meant this would be fixed in Hyrax. However looking at them now, it seems that commits were made in CC, so the before/after indexing behavior may have started with a pre-Hyrax CC upgrade.
[https://github.com/samvera/curation_concerns/issues/1117](https://github.com/samvera/curation_concerns/issues/1117)
[https://github.com/samvera/curation_concerns/issues/559](https://github.com/samvera/curation_concerns/issues/559)

FYI
```
[1] pry(main)> Solrizer.solr_name(:file_size, Solrizer::Descriptor.new(:long, :stored))
=> "file_size_lts"
```

Issue #1055, opened 2017-07-06
Create widgets for Google Analytics - Publisher level


mbakeryo:
Publisher level analytics should be viewable via the dashboard. 

Refer to this usage report as an example of the  metrics to display: https://datastudio.google.com/org//reporting/0B-CNui8jTxj4MHByUVlwRnJfQWc/page/1M

First pass, thinking life to date (for Date Range)is acceptable; add more nuance as we get feedback.

Metrics of interest:
- [x] Users 
- [x] Sessions
- [x] Page views
- [x] Bounce rate
- [x] Top Pages
- [x] Top Landing Pages
- [x] Channels
- [x] Referrals

moseshll:
Am getting data into Rails cache and just displaying raw data for now. Considering next steps for turning some of these into graphs.
<img width="1187" alt="2017-07-25 04 14 14 pm" src="https://user-images.githubusercontent.com/26604206/28591606-55455520-7154-11e7-92e8-002911eeba11.png">

PR #1312, opened 2017-10-26
Addresses #1055 adding Analytics page for platform admins.

moseshll:
This will probably not suffice to close the ticket, since I believe the intent is for press admins to be able to see analytics tailored for them.

Renamed concerns/analytics_presenter.rb to CCAnalyticsPresenter to avoid a naming conflict. That and my AnalyticsPresenter should probably be merged or refactored in some dramatically smarter way.

### DIFF
--- a/app/controllers/analytics_controller.rb
+++ b/app/controllers/analytics_controller.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class AnalyticsController < ApplicationController
+  before_action :authenticate_user!
+
+  def show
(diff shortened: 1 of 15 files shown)


sethaj:
This all seems good. The data is being cached which I think is the most important part here, and none of it looks like it will conflict with hyrax2. We'll probably merge AnalyticsPresenter and CCAnalyitcsPresenter at some point, and of course we'll iterate on how it's all displayed. This gets us to a good place to start iterating.

PR #1368, opened 2017-11-17
Updates to the README to be more community friendly

### LINKED ISSUE #1325: Improve the README for the samvera community
To help the Samvera community better evaluate the Fulcrum suite and heliotrope in particular, 

- [x] include a list of features specific to heliotrope that are not a part of hyrax
- [x] include a mention of cozy-sun-bear and its inclusion as a gem
- [x] adapt or link to samvera contribution guidelines?

jmcglone:
Resolves #1325 

Please read, review and provide feedback on what is missing or could be taken out. FWIW, it is much easier to read at https://github.com/mlibrary/heliotrope/tree/docs/readme

Wiki pages that have been updated in relation to the README:
https://github.com/mlibrary/heliotrope/wiki/EPUB-Reader
https://github.com/mlibrary/heliotrope/wiki/How-to-Contribute

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,25 +19,25 @@ Jump in: [![Stories in Ready](https://badge.waffle.io/mlibrary/heliotrope.png?la
 * [Acknowledgements](#acknowledgements)
 
 # What is Fulcrum? What is Heliotrope?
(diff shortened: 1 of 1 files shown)


conorom:
Very nicely written.
nitpick: I would consider linking the first occurrence of "FileSets" and "Works" to
http://pcdm.org/works#FileSet
and
http://pcdm.org/works#Work

Also, Hyrax have removed the `./bin/bundle exec ./bin/rails hyrax:workflow:load` step from their own README in v2.0. The task is still there but I presume the default workflow gets created some other way now? Don't want to dig into it, TBH.
All's I know is stuff used to break when you didn't run the workflow task and now it doesn't ;-)
Can't go wrong deferring to the change in their README on this one, methinks.

jmcglone:
Thanks @conorom - good points. Added the links you suggested and removed that step. I also moved it to after you start running the application, which I think is right. It somehow got jumbled before running the app when I was working on re-writing things.

Issue #1386, opened 2017-11-30
Penn State banner messed up


conorom:
It only happens when you are logged in and presumably is a post-Hyrax-2 niggle.

I'm guessing this is an easy fix (and for now only visible to us) but perhaps it feeds into a broader discussion of making publisher customization of styles more sustainable (more similar to each other and locked-down, say).

![screen shot 2017-11-30 at 12 05 24 pm](https://user-images.githubusercontent.com/9595080/33444232-b5a0c9c2-d5c7-11e7-9e1b-4b04b9d1fe54.png)

moseshll:
Opinion?
<img width="1215" alt="2017-12-05 03 53 46 pm" src="https://user-images.githubusercontent.com/26604206/33630238-9be2ff94-d9d4-11e7-91ce-d76f65b92351.png">

conorom:
Looks fine to me but I'll defer to @jmcglone. It seems to bring Penn State more in line with the others, too.

PR #1417, opened 2017-12-07
Addresses #1386 messed-up logged-in Penn State banner.

moseshll:
If I had my druthers I would not be using the !important CSS "feature" but oh well.


jmcglone:
Resolves #1386

PR #1436, opened 2017-12-14
Create parent/child self-referential relationship for Press (soon to …

### LINKED ISSUE #1387: Allow publisher to have parent publisher
After discussing the intended use of sub-brands for Gabii, we decided to create Gabii as a publisher and allow this publisher to be a child of the U-M Press publisher. To make this possible, the following will need to happen:

- [x] Add "Parent publisher" field to the Publisher creation web form (eg, https://heliotrope-preview.hydra.lib.umich.edu/presses/new)
- [x] Add a column to the publisher/presses table that can point to the ID of the parent publisher
- [x] Create a mechanism for views to access the parent publisher's data stored in the publisher/presses table.

sethaj:
…be Publisher)

With this PR a child publisher will inherit these field from it's parent if the child doesn't have their own:

footer_block_a
footer_block_b
google_analytics
typekit

Resolves #1387

### DIFF
--- a/app/controllers/presses_controller.rb
+++ b/app/controllers/presses_controller.rb
@@ -30,6 +30,6 @@ class PressesController < ApplicationController
 
     def press_params
       params.require(:press).permit(:subdomain, :name, :logo_path, :description, :press_url, :google_analytics,
-                                    :typekit, :footer_block_a, :footer_block_c, :remove_logo_path)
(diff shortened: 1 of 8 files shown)


sethaj:
In staging I've set gabii to be the child of michigan. I temporarily remove gabii's typekit id and it's using michigan's.

gabii-inherits-michigan-typekit

PR #1438, opened 2017-12-14
Resolves 1383, Allow child publisher to inherit parent publisher theme

### LINKED ISSUE #1383: Allow child publisher to inherit parent publisher theme
Child publishers will need the ability to inherit its parent publisher's theme and database information in views. It will also need to be able to override those things to allow for custom branding at the child publisher level. 

Currently we set a publisher's theme by passing a class to the view. For example in `app/views/press_catalog/index.html.erb` we do:

```
<% provide :page_title, @press.name %>
<% provide :page_class, 'press' %>

<div id="maincontent">
  <%= render 'press_catalog/press_results' %>
```
Which then adds the publisher's subdomain to the `body`, for example:
```
<body class="michigan">
```

For a child publisher to have its parent's and its own theme present, we will need to have both the child's subdomain and the parent publisher's subdomain added to the page. For example:

```
<body class="michigan gabii">
```

- [x] If a child publisher has a parent publisher, add its parent's subdomain across various views
- [ ] The child publisher subdomain should always come second

sethaj:
Resolves #1383

### DIFF
--- a/app/helpers/press_helper.rb
+++ b/app/helpers/press_helper.rb
@@ -12,6 +12,12 @@ module PressHelper
     end
   end
 
+  def press_subdomains(subdomain)
+    press = Press.where(subdomain: subdomain)&.first
+    return if press.blank?
+    [press.subdomain, parent_press(press)&.subdomain].compact
+  end
+
   def name(subdomain)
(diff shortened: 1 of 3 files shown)


sethaj:
This is working for gabii in staging currently.

PR #1469, opened 2018-01-10
store and show monograph handles

### LINKED ISSUE #1460: Monograph handle field
The monograph should have a place to store a handle value. Probably all the same logic as is used for FileSets can apply (HandleService). As in, if it's not set, just assume it's the standard Fulcrum NOID-based handle and show it anyway.

- [ ] add a handle field to Monographs
- [ ] display the field on the monograph catalog, probably at the top near the ISBN values etc.

conorom:
Resolves #1460


jmcglone:
@conorom can you attach a screenshot of the handle display on the monograph page?

conorom:
screen shot 2018-01-10 at 3 24 22 pm

conorom:
@jmcglone It's also on staging, which I was about to tell you in Slack when I was sidetracked by Accesslint Bot.

Issue #1394, opened 2017-11-30
Add additional tabs to the EPUB Monograph catalog page


jmcglone:
If an EPUB Monograph has aboutware, or 3-D models, or databases associated with it, display those as tabs on the Monograph catalog page.

sethaj:
See also #1388

PR #1498, opened 2018-01-12
"featured representatives", tabs for monograph_catalog page

sethaj:
More work toward completing #1394. This provides some infrastructure but the display of the "database" and the "aboutware" is not included.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -15,6 +15,7 @@ Metrics/BlockLength:
 Metrics/ClassLength:
   Exclude:
     - 'app/controllers/catalog_controller.rb'
+    - 'app/presenters/hyrax/file_set_presenter.rb'
 
 Metrics/CyclomaticComplexity:
   Exclude:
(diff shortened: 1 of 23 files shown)


conorom:
Probably worth taking a look at this:
1) upload any new file, even a small image
2) immediately go to the monograph show page (as though to set the new file to be a featured_representative, which is likely to be a fairly common next move)
3) `app/views/featured_representatives/_form.html.erb:20` will fail with an `undefined method 'featured_representatives' for nil:NilClass` because the new file_set has not even been indexed yet
This must mean that the right side of `file_set_presenter.rb:68` can resolve to nil when `file_set_indexer.rb` has not yet set `monograph_id`

I guess the quick thing to do (for now) is a nil check in `app/views/featured_representatives/_form.html.erb:20` which results in, e.g., no drop-down in this scenario.

conorom:
Fix looked good and I couldn't make it crash any more so I'm happy.

PR #1597, opened 2018-03-05
Resolves #1541: Restrict the Gabii EPUB

### LINKED ISSUE #1541: Restrict the Gabii EPUB
- [x] Create a policy for the Gabii Monograph EPUB asset to restrict it only to authorized users (users for whom a license to it has been defined).
- [x] Restrict access to the EPUB asset
- [x] The Monograph (page) and all other assets (including the WebGL) should be public.

gkostin1966:
POLICY:

If an EPUB is "locked" then you must be an "authorized" user to view the EPUB using the reader, otherwise any user may view the EPUB using the reader.

IMPLEMENTATION: (First Iteration)

A "subscriptions" table was added to the database which consists of two columns: "subscriber" and "publication". They are both string fields that will hold "entity ids" such that each subscriber/publication pair is unique a.k.a. a compound key.

An "entity" model has been added to the application which has two attributes: "type" and "identifier". These attributes are strings and the type attribute is currently restricted to be either "email" or "epub". An "entity id" is constructed like so: "#{type}:#{identifier}".

A "subscriber" is uniquely identified by an "entity" of type "email" with the user's email address being the identifier.

A "publication" is uniquely identified by an "entity" of type "epub" with the EPUB's noid being the identifier.

An EPUB is "locked" when there exist a subscription record where the "subscriber" and the "publisher" both equal the EPUB's "entity id". For example an EPUB with noid 999999999 would be lock by row "epub:999999999", "epub:999999999".

A user is "authorized" when there exist a subscription record where the "subscriber" field equals "email:<user's email>" and the "publication" field equals "epub:. For example a user "wolverine@umich.edu" is authorized to read the EPUB "999999999" if the row "email:wolverive@umich.edu", "epub:99999999" exist.

An EPUB must be "locked" for subscription authorization to be enforced. When an EPUB is "unlocked" subscription authorization will not be enforced and any user will be able to read the EPUB regardless if there exist user subscription records for that EPUB in the subscriptions table.

An EPUB can be locked and unlocked by toggling the "Open"/"Restricted" button which is located near the feature representative drop down menu and "Set/Unset" button. When the EPUB is locked the toggle button will display the label "Open" and the "Unset" button will be hidden. Hence, you may only unset a epub feature representative when it is unlocked; the toggle button will display the label "Restrict" and the "Unset" button will be visible.

Currently if you attempt to read a locked EPUB while not logged in you'll be redirected to the login page. If you are logged in but do not have authorization you'll be shown a list of subscriptions for the EPUB beginning with the "lock" and ending with a link that will create a subscription for you in the subscriptions table. Hence after clicking the link the next time you attempt to read the EPUB it will succeed.

One final implementation note. Currently the session variable session[:show_set] is being used to cache the set of epub noids a users may view. I'm not sure how secure this is but I think HTTP Authentication is using a session variable to cache an authenticated user's id. I'm caching the noid to avoid checking authorization for every request for an EPUB's internal files.

I'm not attached to this solution so if anyone has better or different ideas on how to implement this authorization stuff please let me know.

### DIFF
--- a/app/controllers/e_pubs_controller.rb
+++ b/app/controllers/e_pubs_controller.rb
@@ -1,38 +1,37 @@
 # frozen_string_literal: true
 
 class EPubsController < ApplicationController
+  before_action :set_presenter, only: %i[show lock]
+  before_action :set_show, only: %i[show]
+
   def show
-    @presenter = Hyrax::FileSetPresenter.new(SolrDocument.new(FileSet.find(params[:id]).to_solr), current_ability, request)
-    if @presenter.epub?
-      FactoryService.e_pub_publication(params[:id]) # cache epub
-      @title = @presenter.parent.present? ? @presenter.parent.title : @presenter.title
-      @citable_link = @presenter.citable_link
-      @creator_given_name = @presenter.creator_given_name
-      @creator_family_name = @presenter.creator_family_name
-      @back_link = params[:publisher].present? ? URI.join(main_app.root_url, params[:publisher]).to_s : main_app.monograph_catalog_url(@presenter.monograph_id)
-      @subdomain = @presenter.monograph.subdomain
-      @search_url = main_app.epub_search_url(params[:id], q: "").gsub!(/locale=en&/, '')
-      @monograph_presenter = nil
-      if @presenter.parent.present?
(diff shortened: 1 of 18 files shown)


gkostin1966:
Requirements have changed.

PR #1728, opened 2018-04-07
Resolves #1604: Paywall REST API Implementation

### LINKED ISSUE #1604: Paywall REST API Implementation
- [x] Implement Paywall REST API Specification https://tools.lib.umich.edu/confluence/display/FUL/REST+API
https://github.com/mlibrary/heliotrope/tree/product_lease_api

gkostin1966:


gkostin1966:
Fail.  Need to use :id 'cause :identifiers don't work as I expected.

Issue #1692, opened 2018-03-29
FactoryService Cache Fail


gkostin1966:
./app/services/factory_service.rb works in theory but fails in practice.

- [x] Revisit the purpose of the FactoryService
- [x] Redesign the FactoryService cache
- [x] Implement the redesign
- [x] Test the redesign
- [x] Repeat Until Success
- [x] consider caching EPUBs and WebGL at the end of the import process. In many ways unpacking them for display in the app is analogous to derivative creation. Also, triggering this from a user action is dodgy and causes issues, especially when one unpacked item is reliant on another.

conorom:
Added point about unpacking at ingest time not on user action

conorom:
Confirming that after importing 13 EPUBs for HOB only about half would get unpacked and loaded on first browser load. Took one or two page reloads to get the EPUB to display. This will be more of an issue when we import 1000s of EPUBs and it won't be one of us triggering the caching for the first time. The EPUB will just look broken to somebody.

sethaj:
I think we'll probably have to unpack them at "featured representative setting time" rather than ingest. Well for new things. For replacement, we'll need to do it at ingest since it's already an "epub" or "webgl" featured rep.

sethaj:
I think then, that we've decided on this (right?):

1) "Cached" content (epub, webgl) will be considered derivatives ala Hyrax. "Caching" a file had meant just pulling it out of Fedora and putting it on the file system.
We'll keep doing that, but now we'll put them where hyrax puts derivatives `./tmp/derivatives/<pair-tree>`. We're going to call this "unpacking" or an "unpacked" epub/webgl/file and not "caching/cached" from now on.

2) When a FileSet's FeaturedRepresentative is set (epub or webgl), the unpacked file will be created. We do this now and not during ingest because the ingestion process doesn't know the file is a epub/webgl. A user needs to do that (currently).

3) When a FileSet with a FeaturedRepresentative of "epub" or "webgl" is updated ("reversioned" with a new epub/webgl) the new file is unpacked (the old removed first).

But there's a lot of interdependant code right now and it's a little confusing. So to accomplish this, we're going to try to develop in smallish PRs with new code exisiting in parallel to the old code. Once the new code is in place, we'll "convert" all the epubs/webgls to the new "unpacked" model using a rake task. If the new "unpacking" code path works for everyone, we'll start removing the old code and eventually the old "cached" content.

Part 1
[...]

PR #1799, opened 2018-05-04
Use the unpacked derivatives for epubs and webgls if they exist, othe…

sethaj:
…rwise fall back to the "cached" versions.

This is a temporary step towards removing the "cache" completely.

The intention is have both "unpack" and "cache" code paths released into production. The "unpack" code path is triggered if there's an epub or webgl derivative path (an upacked asset in `./tmp/derivatives`). All epubs and webgl derivatives can be created with the `rake heliotrope:unpack_assets` rake task which kicks off UnpackJobs. 

This is currently deployed on both __staging__ and __testing__

Then we'll work on removing the "cache" code from the code base for the next production release.

See #1692

### DIFF
--- a/lib/e_pub/publication.rb
+++ b/lib/e_pub/publication.rb
@@ -42,6 +42,7 @@ module EPub
     end
 
     def self.from_directory(root_path)
+      ::EPub.logger.info("Opening Publication from directory #{root_path}")
       return null_object unless File.exist? root_path
       valid_epub = Validator.from_directory(root_path)
       return null_object if valid_epub.is_a?(ValidatorNullObject)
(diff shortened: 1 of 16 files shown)


lib/e_pub/publication.rb
```diff
@@ -89,10 +90,14 @@ def purge
...
-      entry_file = ::EPub.path_entry(id, file_entry)
+      if @root_path.present?
+        entry_file = File.join(root_path, file_entry)
+      else
+        return Publication.null_object.read(file_entry) unless Cache.cached?(id)
+        entry_file = ::EPub.path_entry(id, file_entry)
+      end
       return Publication.null_object.read(file_entry) unless File.exist?(entry_file)
-      FileUtils.touch(::EPub.path(id)) # Reset the time to live for the entire cached EPUB
+      FileUtils.touch(::EPub.path(id)) # if root_path.blank? # Reset the time to live for the entire cached EPUB
```
conorom: Everything seems to be working smoothly but, even when unpacked the "new" way, this touch of the old path is happening, producing an empty file. Is that intentional, perhaps to continue leaving a record of all epubs in the cache dir? I see there is a commented conditional here (that would touch the new way only).

sethaj: Oops. Yes, thanks for catching this! The conditional should not be commented out (I was debugging and didn't uncomment it.). I'll fix it.

PR #1832, opened 2018-05-22
HELIO-1691 Generalize restricted message (for restricted epubs)

sethaj:


db/schema.rb
```diff
@@ -497,6 +498,14 @@
     t.index ["permission_template_id", "name"], name: "index_sipity_workflows_on_permission_template_and_name", unique: true
   end
 
+  create_table "subscriptions", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
```
conorom: is this related to this work?

sethaj: Ug. No. Thank you. I'll remove it.

PR #1867, opened 2018-06-19
discard creator text after second comma

conorom:


gkostin1966:
Will probably be configurable in the future, but until the use case presents itself this works.

PR #1880, opened 2018-06-26
replace FileSets by name from a directory

conorom:


lib/tasks/replace_files.rake
```diff
@@ -0,0 +1,44 @@
...
+        # Ldp::Gone errors in Hydra::Works::AddFileToFileSet
+        f = FileSet.where(label: new_file).first
+
+        Hydra::Works::AddFileToFileSet.call(f, File.open(new_file_path), :original_file)
+        CharacterizeJob.perform_later(f, f.original_file.id, nil)
+
+        case File.extname(new_file)
+        when 'epub'
+          UnpackJob.perform_later(f.id, 'epub')
+        when 'webgl'
```
sethaj: It looks like the webgl files don't have a `webgl` file extension (or course). The most recent gabii is `AreaBWebGL_Mar29_18_gzip.zip`. So, maybe instead do a FeaturedRepresentative this like:
```
case FeaturedRepresentative.where(file_set_id: f.id).first.kind
when 'epub'
  UnpackJob.perform_later(f.id, 'epub')
when 'webgl'
  UnpackJob.perform_later(f.id, 'webgl')
end
```
or something like that...

conorom: Yes this is less lazy and EPUB-focused. Change made and retested (for EPUBs only, naturellement). Will see what a force-push does to this comment thread.

conorom: Removing this block altogether, as agreed, seeing as CharacterizeJob calls UnpackJob for any FeaturedRepresentative.

PR #1935, opened 2018-08-02
Resolves HELIO-1966: Display complete TOC presenter

gkostin1966:
Refactor of epub presenter, introduction of epub section class, and toc outline view.


gkostin1966:
Deployed to preview for testing first.

PR #1988, opened 2018-09-07
fix monograph single/batch forms

conorom:
Resolves HELIO-2066


app/controllers/hyrax/batch_uploads_controller.rb
```diff
@@ -100,8 +99,7 @@ def attributes_for_actor
       end
 
       def authorize_press
-        curation_concern.press = params[:batch_upload_item][:press]
-        authorize!(:create, curation_concern)
+        authorize!(:create, Monograph.new(press: params[:batch_upload_item][:press]))
```
conorom: @sethaj I'm not crazy about this for some reason but a better way didn't occur to me.

PR #2046, opened 2018-10-12
598 - Network read timeout error a.k.a. fallback page

gkostin1966:
Needed for HELIO-2004


gkostin1966:
Apparently the 503 page is fine for the fallback page so closing without merging this request.

PR #2068, opened 2018-10-24
Resolves 2254: Fulcrum Partials Show Views

gkostin1966:
Now the fulcrum dashboard has index and show pages for products, components, individuals, institutions, publishers (currently not used), and users make for a better user experience IMHO.


gkostin1966:
accesslint doesn't appear to like my unorthodox use of table rows.  It is my preference to ignore accesslint for now since these are platform administrator only views and they are in flux.

PR #2069, opened 2018-10-24
HELIO-2257 COUNTER(like) report rake task for all Institutions and 1 …

sethaj:
…Press


app/services/counter_reporter/title_params.rb
```diff
@@ -83,7 +83,11 @@ def tr_b1
 
       def tr(params)
         @report_title = 'Title Master Report'
-        @metric_types = [params[:metric_type]] || []
+        @metric_types = if params[:metric_type].is_a?(Array)
```
gkostin1966: @metric_types =  [params[:metric_type]].flatten

PR #2100, opened 2018-11-12
Addresses HELIO-2226 Publisher Specific Favicon.

moseshll:
Includes provisional icons for heb/michigan/leverpress, but these should be revisited by a designer. shared/_favicon.html.erb has my notes on getting the icons generated and in the filesystem for display.


moseshll:
Closing this so that we can get better icons on the go for the likes of Lever Press.

PR #2111, opened 2018-11-26
HELIO-1871 COUNTER Item Master Report and Multimedia Report

sethaj:


app/views/counter_reports/without_customer_id/master_view/_form.html.erb
```diff
@@ -23,8 +23,10 @@
         <option value="Unique_Item_Investigations">Unique Item Investiations</option>
         <option value="Unique_Title_Investigations">Unique Title Investigations</option>
         <option value="Total_Item_Requests">Total Item Requests</option>
+        <% unless report_id == 'ir' %>
```
sethaj: There are notes in HELIO-1871 about why this is. We deal with it in a future issue.

PR #2122, opened 2018-11-29
Turn off fallback_to_idp and display :internal_server_error instead

gkostin1966:
Don't mask Shibboleth configuration error, blow up instead!


gkostin1966:
oops

PR #2135, opened 2018-12-11
Resolves HELIO-2308 open access icon.

moseshll:
Also incorporates a late-breaking fix for the monograph model and presenter both calling `.first()` on the `open_access` Solr field.


app/presenters/hyrax/monograph_presenter.rb
```diff
@@ -92,7 +92,7 @@ def holding_contact?
     end
 
     def open_access?
-      solr_document.open_access&.first&.casecmp('yes')&.zero? || false
+      solr_document.open_access&.casecmp('yes')&.zero? || false
```
conorom: This change no longer needed.

PR #2200, opened 2019-01-24
Addresses HELIO-2194 redirect representative file set.

moseshll:


conorom:
@sethaj is correct in that I wrote this ticket to prevent our FeaturedRepresentatives (especially epub, reviews, related) from having their show page indexed by search engines or seen by folks generally, but now that I think about it @moseshll is right too in that covers/thumbnails aren't for consumption in that way either (specifically excluded from our search results etc). I'll amend this commit with a version that combines both.

PR #2217, opened 2019-02-12
aptrust-alpha

gordonleacock:
- Adds: a rake task "bag_given_press[press_id]" at prepares bags for a given press for upload to aptrust in conjunction with the code added and modified in
- Modified: lib/export/exporter.rb, added a new method export_bag and modified method extract to allow a path parameter.


gkostin1966:
Branch no longer active.

PR #2224, opened 2019-02-21
HELIO-2461 fix render time of _index_epub_toc.html.erb

sethaj:
* Do not calculate CFIs for TOC members with anchor tags
* Do not re-open the epub every time monograph_presenter.epub_presenter is called

According to skylight and testing this is a vast improvment, but is still not perfect.

### DIFF
--- a/app/presenters/concerns/featured_representatives/monograph_presenter.rb
+++ b/app/presenters/concerns/featured_representatives/monograph_presenter.rb
@@ -22,7 +22,7 @@ module FeaturedRepresentatives
     end
 
     def epub_presenter
-      EPubPresenter.new(EPub::Publication.from_directory(UnpackService.root_path_from_noid(epub_id, 'epub')))
+      @epub_presenter ||= EPubPresenter.new(EPub::Publication.from_directory(UnpackService.root_path_from_noid(epub_id, 'epub')))
     end
 
     def webgl?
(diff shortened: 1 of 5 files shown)


app/presenters/concerns/featured_representatives/monograph_presenter.rb
```diff
@@ -43,7 +43,7 @@ def database?
 
     def database
```
gkostin1966: return @database if @database.present?

app/presenters/concerns/featured_representatives/monograph_presenter.rb
```diff
@@ -56,7 +56,7 @@ def aboutware?
 
     def aboutware
```
gkostin1966: return @aboutware if @aboutware.present?

app/presenters/concerns/featured_representatives/monograph_presenter.rb
```diff
@@ -115,7 +115,7 @@ def peer_review?
 
     def peer_review
```
gkostin1966: return @peer_review if @peer_review.present?

app/presenters/concerns/featured_representatives/monograph_presenter.rb
```diff
@@ -71,7 +71,7 @@ def reviews
       # This somewhat oddly returns a presenter not a solr_doc. Maybe they all should
       # return presenters?
```
gkostin1966: return @reviews if @reviews.present?

app/presenters/concerns/featured_representatives/monograph_presenter.rb
```diff
@@ -86,7 +86,7 @@ def related
...
```
gkostin1966: return @related if @related.present?

PR #2231, opened 2019-02-25
Greensub Namespace

gkostin1966:
Wrapped Products, Components, Institutions, Individuals, Lessees, and Grants controllers' and views' in Greensub namespace/module.

Primarily replacing model_path with greensub_model_path in views and controllers.


gkostin1966:
Dead End

PR #2279, opened 2019-03-21
2579 page image no chapter downloads

gkostin1966:
heb book
https://heliotrope-staging.hydra.lib.umich.edu/concern/monographs/xg94hp96x?locale=en

heliotrope book
https://heliotrope-staging.hydra.lib.umich.edu/concern/monographs/7w62f8676?locale=en

Open the above links in incognito windows do see the difference between the TOC on different presses.


gkostin1966:
Need to squash it first.

PR #2284, opened 2019-03-22
create_aptrust_logs

gordonleacock:
Not sure - another migration seems to be in here too. Please review carefully!

I seem to be missing timestamps, I thought that was automatic???

I can add a model with filterable too and a controller for it.

Crap - I should have generated a model not a migration!


gordonleacock:
Closed this PR because it wasn't clear it was needed. I also messed it up a bit so starting over (if it is needed) is a better solution.

PR #2337, opened 2019-04-22
2404_aptrust_clean_up4

gordonleacock:
Not longer need to prepend bag names with "fulcrum.org." since fulcrum has its own aptrust domain of that name.
Added 'Internal-Sender-Description' and 'Internal-Sender-Identifier' bag_info.txt since these show up when listed in the Pharos web interface.

### DIFF
--- a/lib/export/exporter.rb
+++ b/lib/export/exporter.rb
@@ -32,9 +32,16 @@ module Export
         return
       end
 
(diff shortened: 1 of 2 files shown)


gordonleacock:
Closed due to an outstanding bug I'm resolving with APTrust staff.

PR #2348, opened 2019-04-26
HELIO-2654 Add a table to track crossref submissions

sethaj:


db/schema.rb
```diff
@@ -697,6 +707,8 @@
     t.index ["work_id"], name: "index_work_view_stats_on_work_id"
   end
 
+  add_foreign_key "api_requests", "users"
+  add_foreign_key "collection_type_participants", "hyrax_collection_types"
```
sethaj: The lines are added back after being removed here: https://github.com/mlibrary/heliotrope/commit/871faf29e56b05655da58eb11cc0e288f17af409#diff-1acd2e7e27a227829d5d14a91c863bb6L702 

At this point I don't know why different developers end up with slightly different schema.db files
after running rake db:migrate. At some point we'll need to try a clean install and try to figure out
what's what. I guess for now these lines can be re-removed from this PR, or we can just merge them
back in. I don't expect serious issues either way...

PR #2354, opened 2019-05-02
HELIO-2655 Add crossref submission api endpoint

sethaj:
Using a token and some xml I'm able to use this to post to test.crossref.org using curl.


gkostin1966:
Where is Crossref::Register defined?

PR #2369, opened 2019-05-13
HELIO-2657 Add a UI that shows the crossref submission log

sethaj:
xref-log-view

Clicking on one of those links will show the file (which is too big to display in the table)

xref-log-file-view


sethaj:
There are some security implications here that I need to think about more.

PR #2396, opened 2019-06-14
Addresses HELIO-2767 Sitemaps should not have epub urls that are rest…

moseshll:
…ricted

### DIFF
--- a/config/sitemap.rb
+++ b/config/sitemap.rb
@@ -48,6 +48,7 @@ SitemapGenerator::Sitemap.create do
       # to their monograph_catalog page so don't need sitemap representation.
       # However, eventually pdf_ebooks will need their own url too. I think.
       next if rep&.kind.present? && rep&.kind != 'epub'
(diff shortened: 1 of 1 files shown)


moseshll:
I was under the impression that the line (51)…
next if rep&.kind.present? && rep&.kind != 'epub'
... would bleed off most of the uninteresting file_sets before the call to EPubPolicy.new(…)
Or is it likely that in most cases rep.kind will be blank?

sethaj:
I think that most FileSets actually won't be "FeaturedRepresentatives" at all, so they'll make it by line 51. The whole "FeaturedRepresentative" is a bad abstraction (I know this because I created it :disappointed:). Most FileSets are not "special" or "FeaturedRepresentatives", so that yes, "kind" will usually be blank or most likely nil since "rep" won't usually exist for a given FileSet.

moseshll:
Opened a "part_deux" pull request bleeding off the file_sets based on MIME type. Adventures in git directly in github. Not sure how I feel about that.

sethaj:
See https://github.com/mlibrary/heliotrope/pull/2403

PR #2400, opened 2019-06-18
HELIO-2659 Add hook to auto-register FileSet DOIs sometimes

sethaj:
1. If the Press is set up to auto-generate DOIs.
2. If we are going from Private to Public
3. If we are *not* updating/adding any file_sets
4. If the Monograph already has a DOI
Then: create DOIs for the FileSet

Does the above seem ok? We only want to trigger this under very specific circumstances.

This also removed the actor_insert code out of config/initializers due to autoload not working right in dev, see https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#autoloading-and-initializers

Tested this on staging (the branch is there now) and the actors load correctly and crossref submission happens successfully (if the Press is set to, if there's a monograph.doi, if there are no file_set create/updates and if the monograph is going from Draft to Public (ie: is being "Published")).

There's no crossref polling set up yet, that's HELIO-2768


spec/config/initializers/hyrax_actor_factory_spec.rb
```diff
@@ -9,28 +9,29 @@
     subject(:middlewares) { actor_factory.middlewares }
 
     it do
-      expect(middlewares.count).to eq 20
+      expect(middlewares.count).to eq 21
```
sethaj: This spec is fine, but it needs to be moved since the file config/initializers/hyrax_actor_factory.rb has been removed and the code put into config/application. Not a big deal, but should be done at some point.

PR #2402, opened 2019-06-18
Update sitemap.rb

moseshll:


sethaj:
See https://github.com/mlibrary/heliotrope/pull/2403

PR #2418, opened 2019-07-10
Addresses HELIO-2809 Generate TOC from representative PDF's outline

moseshll:
There is one major flaw with this implementation (that I know of, anyway LOL): the use of epub_path() in _index_epub_toc.html.erb for ToC links into the PDF reader. They will need to be fixed when we know what our PDF reader URLs are going to look like.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -129,6 +129,9 @@ gem "loofah", ">= 2.2.3"
 # Use MySQL as the database for Active Record
 gem 'mysql2', '~> 0.4.10'
 
+# Read PDF ToC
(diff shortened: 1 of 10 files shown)


gkostin1966:
Please add new specs:

./lib/spec/pdf_ebook/interval_spec.rb
./lib/spec/pdf_ebook/publication_spec.rb
./lib/spec/pdf_ebook_spec.rb

Duplicating the ./lib/spec/e_pub specs of the same name followed by a find and replace with removal of irrelevant test should do the trick.

PR #2429, opened 2019-07-24
Resolves HELIO-2797: Ingest and display a Leaflet map package

gkostin1966:
First Iteration


gkostin1966:
I need to rethink using FeaturedRepresentatives.

PR #2470, opened 2019-09-05
Resolves HELIO-2820: Allow enlarging of cover thumbnails for Monographs

gkostin1966:
modal

### DIFF
--- a/app/assets/stylesheets/application/heliotrope.scss
+++ b/app/assets/stylesheets/application/heliotrope.scss
@@ -449,6 +449,7 @@ footer.press {
 
   .monograph-cover {
     .modal {
+      text-align: center;
       img {
         max-width: 100%;
         max-height: 100%;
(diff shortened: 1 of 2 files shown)


jmcglone:
@gkostin1966 Ok, some suggestions before this gets merged:

1) It is generally expected that a modal window is centered on pages. You can do this by adding a `text-align: center;` to `.modal` in `heliotrope.scss`.
2) Add an `aria-label="Select to enlarge cover"` to the `<button>` element so that it is clear what will happen if selected.
3) I know accesslint failed the lack of `alt` attribute, but now assistive technology is reading alt text twice. I think it would be better to leave the alt attribute but make it empty. So something like `alt=""` would do.
4) On `<div ... class="modal" ...> the `aria-labelledby` attribute points to `modalImageLabel` ID, but that doesn't seem to exist anywhere. My vote would be to remove that and replace it with `aria-label="Select the escape key to close the modal window" so that assistive tech users know how to exit out of the modal window.

Let me know if you have any questions and thanks for working on this!

PR #2471, opened 2019-09-05
Resolves HELIO-2907: PDF Viewer - Rename 'Document Outline' data-l10n-id

gkostin1966:
Because four flavors of en- should be enough.

bundle exec rails jekyll:deploy

Clear browsing data


respinos:
These are modifying the PDF Viewer files, which should be managed in the pdf.js/heliotrope branch. BUT WE SHOULD JUST PUNT ON THESE ATTRIBUTES!!!!

respinos:
Since we're not using icons, we definitely do not need the `title` attributes on these buttons.

gkostin1966:
Overruled ... purge data-l10n-id attribute.

PR #2474, opened 2019-09-06
HELIO-2781 Generate Score work type for carillon

sethaj:
Also HELIO-2912, Allow "score" type fields on FileSets as json
Wireframes: https://tools.lib.umich.edu/jira/browse/HELIO-2522


app/controllers/press_catalog_controller.rb
```diff
@@ -52,24 +52,44 @@ def open_monographs
...
       end
+    end
 
+    def musical_score
+      blacklight_config.add_facet_field Solrizer.solr_name('creator', :facetable), label: 'Composer'
+      blacklight_config.add_facet_field Solrizer.solr_name('octave_compass', :facetable), label: 'Octave Compass'
+      blacklight_config.add_facet_field Solrizer.solr_name('bass_bells_required', :facetable), label: 'Bass Bells Required'
+      blacklight_config.add_facet_field Solrizer.solr_name('solo', :facetable), label: 'Solo'
+      blacklight_config.add_facet_field Solrizer.solr_name('musical_presentation', :facetable), label: 'Musical Presentation'
+      blacklight_config.add_facet_field Solrizer.solr_name('composer_gender', :facetable), label: 'Composer Gender'
```
sethaj: composer_gender and composer_ethnicity should be combined into the single composer_diversity (as they are in the model, presenter and form)

gkostin1966: Next iteration this will be fixed

PR #2480, opened 2019-09-10
Ensure that the content attributes are encoded when building html

bertrama:


bertrama:
See https://www.fulcrum.org/concern/monographs/h128ng045?locale=en for an example.

```
 <meta name="citation_title" content=""I saw a nightmare-- ": doing violence to memory : the Soweto uprising, June 16, 1976">

```

PR #2481, opened 2019-09-10
Resolves HELIO-1829: Add Leaflet full-screen mode

gkostin1966:
Since there is no 'official' convention for naming Leaflet plugins I'd like to start a convention of storing the unzipped downloaded release folder under 'vendor' and copying the necessary files to their appropriate locations under 'vendor/assets'.  This way the next developer will know what version of the plugin is being used as well as what files belong to which plugin.


jmcglone:
@gkostin1966 some notes from testing:

1) Is there supposed to be an icon for the fullscreen button? I don't see anything in FF or Safari. See screenshot from my dev space:

Screen Shot 2019-09-10 at 3 50 42 PM

2) The fullscreen mode does not keep the focus of the keyboard, and it should. For example, when I use my keyboard controls in fullscreen mode to tab advance, I toggle through all the elements of the page, instead of keeping the keyboard's focus only on the elements within the Leaflet image viewer.

PR #2485, opened 2019-09-12
Resolves HELIO-2453: Add FileSet actions to Monograph catalog page

gkostin1966:
Probably the best thing to do after all that is to make sub-tasks to move other elements from the show page if they are useful (like the "Publish Monograph" button). I don't think we need to rush to do this. We can discuss.


gkostin1966:
Feature Branch

PR #2498, opened 2019-09-19
HELIO-2949 Save GA data to the database, not cache

sethaj:
this is on preview now. I've run the add_ga_history.rake task and it's matching up with what we have in the cache.
```
irb(main):011:0> Rails.cache.read('ga_pageviews')['3n203z11f'].map{|date, pageviews| pageviews }.sum
=> 180
irb(main):012:0> GoogleAnalyticsHistory.where(noid: '3n203z11f').map(&:pageviews).sum
  GoogleAnalyticsHistory Load (0.7ms)  SELECT `google_analytics_histories`.* FROM `google_analytics_histories` WHERE `google_analytics_histories`.`noid` = '3n203z11f'
=> 180
irb(main):013:0> 
```
I've added a nightly cron to add yesterdays stats. Everything seems to be right....

### DIFF
--- a/app/presenters/concerns/analytics_presenter.rb
+++ b/app/presenters/concerns/analytics_presenter.rb
@@ -12,26 +12,19 @@ module AnalyticsPresenter
   def timestamped_pageviews_by_ids(ids)
     # we want the output here to be structured for Flot as used in Hyrax, with unixtimes in milliseconds
     ids = ids_array(ids)
-    start_time = date_uploaded.strftime('%Q').to_i
-    pageviews = Rails.cache.read('ga_pageviews')
-    if pageviews.nil?
-      @pageviews ||= '?'
+    data = {}
+    count = 0
+    GoogleAnalyticsHistory.where(noid: ids).each do |r|
+      timestamp = Date.strptime(r.original_date, '%Y%m%d').strftime('%Q').to_i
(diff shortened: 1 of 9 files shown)


app/presenters/concerns/analytics_presenter.rb
```diff
@@ -12,26 +12,19 @@ def ids_array(ids)
...
-    end
-
     data = {}
     count = 0
-    if pageviews.is_a? Hash
-      ids.each do |id|
-        next if pageviews[id].blank?
-        pageviews[id].each do |date, views|
-          timestamp = Date.strptime(date, '%Y%m%d').strftime('%Q').to_i
-          next unless timestamp >= start_time
```
sethaj: We used to do this, to make sure not to include any GA stats that happened before the Work/FileSet date_uploaded. But why? How could there be any usages stats from before a work/file_set was created? Anyway, I stopped doing this check in this new version. I did leave date_uploaded in the #flot_pageviews_over_time method because it's needed for the graph.

conorom: Yeah I remember thinking that over a couple years ago. NOID reuse is one thing I guess. Maybe migration from another app instance. Stuff like that. Doesn't apply to us anyway.

PR #2500, opened 2019-09-20
2930 rubocop oddity

gkostin1966:
Upgrade to latest rubocop gems.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -24,6 +24,12 @@ AllCops:
 Layout/IndentationConsistency:
   EnforcedStyle: indented_internal_methods
 
+Layout/EmptyLinesAroundAccessModifier:
+  EnforcedStyle: around
+
+Layout/EndAlignment:
+  EnforcedStyleAlignWith: keyword
+
 Lint/DuplicateMethods:
   Exclude:
   - 'app/presenters/e_pub_download_presenter.rb'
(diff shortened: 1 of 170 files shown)


conorom:
Really not a fan of the (presumably automatic) alignment corrections of assignment statements using an if block that have been made here. Much nicer and more Ruby-looking to have the if-else-end aligned, IMO.
There is a discussion here mentioning this stuff: https://github.com/rubocop-hq/rubocop/issues/1323
Basically need to set `Lint/EndAlignment` to `keyword`, though I haven't checkout out your branch to see if these "corrections" are just what the autocorrect is doing because that is not set, or if having these laid out in the "more Rubyish" way now breaks rules. Tagging @sethaj to get his $0.02. Note that a commenter on the rubocop-hq project above links some classic Ruby book from 2001 to make the point that aligning the keywords in this scenario is a very Ruby thing to do.

sethaj:
Yeah, that was my only concern about this, the if/else alignment stuff. It isn't really that big of a deal, and we use rubocop to enforce styles so we don't ever have to get into the weeds about exactly this kind of thing, but maybe this time since it is a bit of a departure we can all vote or something and then just make the change (or not) and move forward. Let's talk about it and decide at our next meeting.

gkostin1966:
See squashed PR

PR #2517, opened 2019-10-02
Resolves HELIO-1793 Monograph catalog modals should keep focus until …

gkostin1966:
…closed

modal.js should handle all modals in heliotrope.


gkostin1966:
Oops, still have add ARIA attributes.

PR #2523, opened 2019-10-10
Addresses HELIO-3000 Include Parent Press GA ID on Sub Press pages.

moseshll:
Add new helper method all_ga_analytics to return GA ids as an array (almost always an array of one but an array of two for Gabii) for the GA JS to iterate on.


moseshll:
Closing this PR because the modified code creates two press trackers with the same name ('pressTracker') which probably muddies the waters with ga_event_tracking.js.

PR #2524, opened 2019-10-10
Addresses HELIO-2971 Book cover buttons should not have color backgro…

moseshll:
…unds


moseshll:
No hover selectors appear to be needed, based on testing in Firefox and Safari.

PR #2530, opened 2019-10-16
wait an extra 3 seconds for PDF load

conorom:


conorom:
At best I've made this happen less often. We need to do something other than allow Capybara to wait longer. Closing the PR.

PR #2532, opened 2019-10-18
Monograph covers use PNG thumbnails

conorom:
HELIO-3045

### DIFF
--- a/app/models/riiif/file.rb
+++ b/app/models/riiif/file.rb
@@ -57,8 +57,10 @@ module Riiif
         command << "pnmscalefixed -ysize #{h}"       if h.present? && w.blank?
       end
 
-      # ignore quality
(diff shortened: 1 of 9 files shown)


app/models/riiif/file.rb
```diff
@@ -57,8 +57,8 @@ def extract_tifftopnm(options)
         command << "pnmscalefixed -ysize #{h}"       if h.present? && w.blank?
       end
 
-      # ignore quality
-      command << "pnmtojpeg -quality 95"
+      # ignoring options[:quality] for jpeg here
```
respinos: Technically the "quality" the original comment refers to is the IIIF Image API "quality" (https://iiif.io/api/image/2.1/#quality) which has nothing to do with JPEG quality. We're only generate the "default" quality, not letting the URL specify color/grayscale/etc.

PR #2533, opened 2019-10-23
Resolves HELIO-2953 Books on publisher catalog should be combined int…

gkostin1966:
…o one link

plus refactor.


jmcglone:
@gkostin1966 

In testing with a screen reader, I noticed that the title of the book is read twice: once for the cover image alt text, and again for the `<h3>` which contains the title text. It also is not clear that the "author" text is the author/editor. 

To ensure the title is read only once and to give context to the author/editor name, do the following:

1) leave the cover image `alt` attribute empty. Example:

``

2) add an aria-label to the `<p>` element that wraps the author/editor. Example:

`<p aria-label="By <%= presenter.authors %>" class="authors"><%= presenter.authors %></p>`

PR #2567, opened 2019-11-25
reader uses cached PDF file

conorom:
HELIO-2940


conorom:
Just noticed the old branch I had been sprucing up was still calling for JPGs from RIIIF. Will need to fix that and reopen the PR.

PR #2599, opened 2020-01-10
HELIO-3052 make fewer solr calls

sethaj:
The reason for this branch was from some excessive solr calls for cache key generation for IIIF tiling as talked about in a comment in HELIO-3052 that we there to make sure re-versioned/replaced FileSets were getting new/updated tiles. That works now with md5, not solr. This branch is on staging and I've tested re-versioning image FileSets and IIIF/Leaflet a few times.

Other then that some work was done in "Sighrax" in consultation with Greg before he went on vacation to reduce the number of (redundant) solr calls. And then a few bits of removing other unnecessary solr calls in a few other places.

In the end this branch shouldn't change any functionality but has the potential to greatly reduce the number of solr calls pages make.

### DIFF
--- a/config/initializers/riiif_initializer.rb
+++ b/config/initializers/riiif_initializer.rb
@@ -13,6 +13,7 @@ end
 # each image. If you are using hydra-file_characterization, you have the height & width
 # cached in Solr. The following block directs the info_service to return those values:
 Riiif::Image.info_service = lambda do |id, _file|
+  Rails.logger.debug("[RIIIF] H/W CHECK FOR #{id}")
   doc = ActiveFedora::SolrService.query("{!terms f=id}#{id}", rows: 1).first
   { height: doc["height_is"], width: doc["width_is"] }
 end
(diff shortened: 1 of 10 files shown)


config/initializers/riiif_initializer.rb
```diff
@@ -13,15 +13,18 @@
 # each image. If you are using hydra-file_characterization, you have the height & width
 # cached in Solr. The following block directs the info_service to return those values:
 Riiif::Image.info_service = lambda do |id, _file|
+  Rails.logger.debug("[RIIIF] H/W CHECK FOR #{id}")
```
conorom: @sethaj I can see some value in the logging statement below, but this one will run for every load of a Leaflet map and doesn't seem very informative?

sethaj: I'll remove it. It was just there to see how often it was actually called and was hitting solr.

conorom: Ah, debug so not in production <facepalm>. I'm just going to merge this!

PR #2600, opened 2020-01-14
Update to bundler 2

sethaj:
HELIO-3121


conorom:
Merging in spite of unrelated system spec timing failures.

PR #2629, opened 2020-02-10
Add presses and change name of task to "fulcrum_published_dois"

sethaj:
HELIO-3118
* ran this on preview, works are expected
* once this is merged some crons need to be updated, don't close the ticket.


lib/tasks/handles/fulcrum_published_dois.rake
```diff
@@ -0,0 +1,43 @@
...
+      children << Press.where(subdomain: subdomain).first&.children&.pluck(:subdomain)
+    end
+
+    subdomains = subdomains + children
+    subdomains = subdomains.flatten.compact.uniq
+
+    hits = ActiveFedora::SolrService.query("{!terms f=press_sim}#{subdomains.map(&:downcase).join(',')}", rows: 100_000)
+
+    hits.each do |hit|
+      next unless hit["suppressed_bsi"] == false && hit["visibility_ssi"] == "open"
```
conorom: Not really a nitpick but I'm always leery of this `suppressed_bsi` thing. I'm pretty sure action is only ever taken on this in Hyrax/stack code when it's true (Thus removed from results). I have sort of a vague memory of seeing it missing from Solr docs in the past, but I may have made this up. Is that possible?
Either way, I'd like if this read `hit["suppressed_bsi"].blank?` but would welcome correction or discussion.

gkostin1966: Sounds good to me.  Should added it to rubocop.

gkostin1966: Also. I would prefer /open/i.match?(hit["visibility_ssi"]), love those case insensitive regular expressions.

conorom: thinkin-face

PR #2649, opened 2020-02-26
add clamby instead of clamav for virus scanning

sethaj:
HELIO-3230

### DIFF
--- a/config/initializers/clamav.rb
+++ b/config/initializers/clamav.rb
@@ -1,3 +1,31 @@
 # frozen_string_literal: true
 
-ClamAV.instance.loaddb if defined? ClamAV
+# ClamAV.instance.loaddb if defined? ClamAV
+
+# We're overloading this to use the clamby gem instead of clamav gem
+# I think probably Hyrax will swtich to clamby in version 3.0 but I'm not clear
+# on it. We're doing it now (in 2.7.0).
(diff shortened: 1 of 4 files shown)


config/initializers/clamav.rb
```diff
@@ -1,3 +1,31 @@
...
+# I think probably Hyrax will swtich to clamby in version 3.0 but I'm not clear
+# on it. We're doing it now (in 2.7.0).
+# See https://github.com/samvera/hydra-works/blob/master/lib/hydra/works/virus_scanner.rb
+# and HELIO-3230
+
+if defined?(Clamby)
+  # https://github.com/kobaltz/clamby#configuration
+  Clamby.configure(
+    check: false,
+    # daemonize: true,
```
conorom: Would be preferable to daemonize, right?
All the custom stuff DBD did for a/v scanning was simply to use clamd. If we can get that to work just with this flag then that would be awesome. I just verified clamd is running on both production servers.

sethaj: I mean, yes it would, but there's an error when I tried it. It's permissions related and I'm not sure it's worth the trouble to dig into. It's in the last comment in the ticket. I decided to fight that battle another day since speed isn't really that big of a deal to us... but we could try to figure it out now too.

conorom: Blarg. Should have read the ticket. Should have guessed you'd done your due diligence. Non-daemonized it is. Which is the same as now. Mergin'

PR #2651, opened 2020-02-27
Partially resolves HELIO-478 Use handle service to create handles on creation

gkostin1966:
This pull request is mainly plumbing.  The rake task /lib/tasks/handle.rake will be run daily by cron.  Focus on understanding the HandleJob, all the other code supports this job.  I created a new table HandleDeposits and the model, view, and controller to support how it is being used.  I also added a new job form to the /fulcrum/jobs page (third one down) to create a handle for a noid a.k.a. Monograph, Score, or FileSet.  Go ahead and create a handle for a Monograph since the HandleService update and delete methods have yet to be implemented so no handles will be created or deleted.  The /handle_deposits page will display the HandleDeposits table and you'll see your request to create a handle for the Monograph has a verified status of false.

If you have any questions you know where to find me.

### DIFF
--- a/app/controllers/fulcrum_controller.rb
+++ b/app/controllers/fulcrum_controller.rb
@@ -11,12 +11,21 @@ class FulcrumController < ApplicationController
     case params[:cmd]
     when 'ingest'
       ExtractIngestJob.perform_later(params[:token], params[:base], params[:source], params[:target])
+    when 'unpack'
+      UnpackJob.perform_now(params[:noid], params[:kind])
+    when 'handle'
+      case params[:job]
+      when 'create'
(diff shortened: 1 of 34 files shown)


sethaj:
I'm still absorbing this, but I wonder if it would be possible to put all the new handle jobs into their own directory: app/jobs/handles/ ? The jobs are starting to multiply and it might be nice to organize them a bit.

gkostin1966:
If creating a subfolder "handles" under the jobs folder requires place the handle jobs inside a handles module, do you still want to do it?

gkostin1966:
I reworked HandleJob and added better comments, give it another read it should make more sense.

PR #2683, opened 2020-03-12
HELIO-3269 Set BAR publisher page to default to List View.

moseshll:


db/schema.rb
```diff
@@ -356,6 +360,19 @@
     t.index ["source_id"], name: "index_permission_templates_on_source_id", unique: true
   end
 
+  create_table "permits", id: :integer, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
```
gkostin1966: I think the db migrations are messed up on your development machine.

moseshll: I'm gonna delete this PR and rebuild my heliotrope dir from scratch. Everything is hosed now.

PR #2688, opened 2020-03-13
fulcrum jobs page cleanup and removal of devproque environment

gkostin1966:
@conorom fixed something and now the devproque environment is no longer needed.


gkostin1966:
Retract

PR #2702, opened 2020-03-19
HELIO-3262 [Firefox] PDF EReader Preferences Button Does Not Open

moseshll:
Asynchronize the code that deals with the zoom level radio buttons in the prefs modal, so the radio buttons actually exist before futzing with them.


respinos:
`async` isn't in IE11 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function (barring we haven't pulled in a polyfill somewhere)

moseshll:
Closing this in favor of a considerably less slick solution.

PR #2710, opened 2020-03-24
HELIO-3245 Remove File Manager Radio Buttons for Non-Image FileSets

moseshll:


app/presenters/hyrax/file_set_presenter.rb
```diff
@@ -303,6 +303,14 @@ def score_version
       solr_document['score_version_tesim']
     end
 
+    # Returns true if the MIME type is image, or if MIME type is not available,
+    # if the file extension is a known image type.
+    def probable_image?
```
conorom: Hate to do this but I'm wondering if we need this method. We use this one in many places already: https://github.com/samvera/hydra-works/blob/74227817c04d916b9ab84f247440606289cc8eea/lib/hydra/works/models/concerns/file_set/mime_types.rb#L10

moseshll: I think this particular brand of "look for image MIME type and failing that look at file extension" approach is unique to this scenario. Thus far anyway. And I think matching just on "image/" rather than expecting Hyrax to be be correct, is appropriately paranoid.

I could swap in `Hydra::Works::MimeTypes.image_mime_types.member? mime_type` on line 309 but we still need the extension check.

PR #2720, opened 2020-03-27
Patch ActionView for HELIO-3304

sethaj:


gkostin1966:
How do I going about testing this?  Just deploy to staging and see if it works?

sethaj:
Good point, sorry it slipped my mind. Let me come up with something and I'll add it to the PR or ticket.

PR #2722, opened 2020-03-30
HELIO-3277 Don't parse PDF ToC on every pdf_ebook reader Monograph ca…

moseshll:
…talog load


sethaj:
This is great, but I think I'd like it to include epubs as well as pdf_ebooks. So I'm going to close this, edit the original ticket, take over this branch and see if we can add that.

PR #2723, opened 2020-03-30
Resolves HELIO-3259 FeaturedRepresentative unique index not working

gkostin1966:
FYI: Ran the following rake task on production and it came up empty, hence the database migration should succeed when released to production.

desc 'Find non unique kind works'
namespace :heliotrope do
  task find_non_unique_kind_works: :environment do
    FeaturedRepresentative.distinct.pluck(:work_id).each do |work_id|
      kind_count = FeaturedRepresentative.where(work_id: work_id).count
      unique_kind_count = FeaturedRepresentative.where(work_id: work_id).distinct.pluck(:kind).count
      next if kind_count == unique_kind_count
      puts Sighrax.url(Sighrax.from_noid(work_id))
    end
  end
end

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20200313153535) do
+ActiveRecord::Schema.define(version: 20200330165632) do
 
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -381,6 +382,7 @@
     t.text "navigation_block"
     t.boolean "default_list_view", default: false
     t.boolean "aboutware", default: false
+    t.boolean "restricted", default: true
```
sethaj: This is maybe not supposed to be here...?

PR #2740, opened 2020-04-13
Resolves HELIO-3075 and HELIO-3076 Add visual descriptions, captions,…

gkostin1966:
… and refactor transcript FileSet metadata.

Going to use HELIO-746 to fix ableplayer behavior.


sethaj:
Instead of "captions" would it make sense to name them "closed_captions" and instead of "descriptions", "visual_descriptions"? Just wondering if it might be less confusing since we already have a singular "caption" which is not an a11y closed caption, and a singular "description" which is not an a11y visual description.

gkostin1966:
I initially had named them "closed_captions" and "visual_descriptions" but then I looked at the video tag in the view:

```
  <% if file_set.captions.present? %><%# kind="subtitles" not "captions" due to https://github.com/mlibrary/heliotrope/issues/1234 %>
    <track kind="subtitles" src="<%= hyrax.download_path(file_set, file: 'captions_vtt') %>" srclang="en" label="English"  />
  <% end %>
  <% if file_set.descriptions.present? %>
    <track kind="descriptions" src="<%= hyrax.download_path(file_set, file: 'descriptions_vtt') %>" srclang="en" label="English"  />
  <% end %>

```
and changed them to match the track kind.   I have no issue renaming them as you suggest, perhaps @conorom or @jmcglone could add their two coppers.

conorom:
I guess using the more descriptive field names makes things clearer within our ever-growing set of metadata. Hey, maybe `textual_description` and `closed_caption`, even? Keep 'em singular.
Or is there a chance we're going to try to use these multi-valued at some point, with a magic comment inside to indicate, e.g. language?
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/track

gkostin1966:
There is mention of multi language support in other tickets and translations is a multi valued field and this ticket mentions translations as an example.  Current the language for captions and descriptions is hard coded to be english in the view.

sethaj:
Ug, yeah, that kind="descriptions" or kind="captions" markup.... right. I guess I'd personally prefer that we be more specific on the backend so the Work/hyrax metadata would be "visual_descriptions" and "closed_captions" to keep things clearer for us. 

And then in the view we use whatever the browsers tell us to do for the specialized player markup. IDK...

gkostin1966:
Changed metadata names to closed_captions and visual_descriptions.

PR #2741, opened 2020-04-13
open_access should only be 'yes'

conorom:
HELIO-3287


conorom:
Reopening to get AccessLint to rereun. It's wasted a bunch of my time trying to figure out what to do with a hidden field that has no label. Looks like this is acknowledged as a bug here:
https://github.com/AccessLint/accesslint.js/issues/32

PR #2743, opened 2020-04-14
Quick fix to migrate transcript job to process videos only.

gkostin1966:
Migrate transcript to closed captions of video file sets only.


conorom:
Any reason not to do `file_set.transcript = nil` before the save as well?
It is expected to be unused after this right? Until we maybe come up with a new use for that field.

gkostin1966:
The transcript needs to exist in order to copy it into the closed_captions field.

conorom:
Ah, but I meant after that in the task. But pursuant to Sprint Planning discussion we can see what happens with the transcript field display in the next few sprints and decide then. Merging.

PR #2762, opened 2020-05-01
don't use parent's press_statistics GA URLs

conorom:
HELIO-3362


conorom:
closed in favor of hot fix

PR #2776, opened 2020-05-20
move stats graph calculations to a service

conorom:
HELIO-3331


conorom:
This is on preview now. I've checked quite a few FileSet and Monograph stats graphs and they look sensible. Verified a few against their upload date from the console as well.

PR #2807, opened 2020-07-15
service for MIME type overrides, add 'video/mpg'

conorom:
HELIO-3438


conorom:
typo in commit message!

PR #2826, opened 2020-08-25
allow hyphens in subdomains

conorom:
HELIO-3509


gkostin1966:
Do we care if a----------------------------------b is valid?

conorom:
Haha. That's pretty annoying. I was focused on (and struggling with) what was actually in the ticket. Given that it's in-house people you'd think we'd be OK without such thorough validation but I think I might have to obsess about and fix that. Probably by abandoning Regex in favor of a [custom validator method](https://guides.rubyonrails.org/active_record_validations.html#performing-custom-validations) or at least combining it with one.

PR #2830, opened 2020-08-26
don't cache institutions in session variable

conorom:
HELIO-3488


gkostin1966:
We should probably use a local cache though...

```  
  def current_institutions?
    current_institutions.count.positive?
  end

  def current_institutions
    @current_institutions = (ip_based_institutions + shib_institutions).uniq
  end
```

Just saying.

conorom:
Yup, that'll at least stop it from happening twice per page load, looks like.

PR #2845, opened 2020-09-14
bump toc to mediumtext

conorom:
HELIO-3549


conorom:
@sethaj I presume there are no hidden dangers with the generate having put `ActiveRecord::Migration[5.2]` in the migration?
Let me know if you think it's dodgy and should be edited to 5.1 for whatever bizarre reason.

PR #2847, opened 2020-09-16
output CSV for all of a press's monographs

conorom:
HELIO-2536


conorom:
I'm PR'ing this task in its clean "using the exporter as-is" state. I'll leave the ticket open because we're probably going to add technical metadata to this exported CSV, somehow, soonish.

PR #2853, opened 2020-09-21
OPDS Catalog 2.0

gkostin1966:
Resolves HELIO-3483 Create OPDS feed for UMPEBC OA titles in SimplyE

### DIFF
--- a/app/models/opds/publication.rb
+++ b/app/models/opds/publication.rb
@@ -1,0 +1,249 @@
+# frozen_string_literal: true
+
+module Opds
+  class Publication
+    private_class_method :new
+
+    delegate :to_json, to: :to_h
+
+    class << self
+      def new_from_monograph(monograph)
+        Publication.send(:new, monograph)
+      end
+    end
+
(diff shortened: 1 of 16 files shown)


app/models/opds/publication.rb
```diff
@@ -0,0 +1,248 @@
...
+      def title_sort_as
+      end
+
+      def download_ebook_url(ebook)
+        Rails.application.routes.url_helpers.download_ebook_url(ebook.noid)
+      end
+
+      def monograph_image_url(monograph, width_size = nil)
+        width_size ||= 'full'
+        Riiif::Engine.routes.url_helpers.image_url(monograph.noid, host: Rails.application.routes.url_helpers.root_url, size: width_size, format: 'jpg')
```
sethaj, line 241: We'll need to use the cover (file_set) noid here instead of the monographs noid (see note in HELIO-3483)

gkostin1966: Switched to cover for images.

PR #2858, opened 2020-10-02
search/filter within press hides jumbotron

conorom:
HELIO-3477

### DIFF
--- a/app/controllers/press_catalog_controller.rb
+++ b/app/controllers/press_catalog_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class PressCatalogController < ::CatalogController
-  before_action :load_press
+  before_action :load_press, :search_ongoing
   before_action :conditional_blacklight_configuration
 
   self.theme = 'hyrax'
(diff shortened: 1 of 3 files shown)


app/controllers/press_catalog_controller.rb
```diff
@@ -49,6 +49,18 @@ def load_press
       render file: Rails.root.join('public', '404.html'), status: :not_found, layout: false
     end
 
+    def search_ongoing
+      @search_ongoing = false
+
+      params.each do |key, _value|
+        non_search_keys = ['action', 'authenticity_token', 'controller', 'locale', 'page', 'per_page', 'press', 'sort', 'utf8', 'view']
```
gkostin1966, line 56: Brittle logic but I can't think of an alternative.

conorom: Yes, it's totally possible that other params will need to be added after running/testing this for a while.
We could reverse the logic and actually store all the Solr field names for facets instead? I can go that way if you think it's worth it. Later could add human-readable names in such a list as well, to use in a very elaborate message for https://tools.lib.umich.edu/jira/browse/HELIO-3478
I was planning on avoiding both of those things because it seems like adding maintenance nightmares on top of the feature creep. Dunno.

gkostin1966: Yeah, I was thinking the search_keys would be a much bigger and uglier set hence I can stomach the brittle in this context.

conorom: OK resolved in parking lot. Good enough for now.

PR #2863, opened 2020-10-09
HELIO-3320 - Updates to fulcrum homepage

jmcglone:
Update fulcrum homepage with:

new partners
new case studies
new case study page
updates to partner with us page


conorom:
@jmcglone hovering over the new partners' icons, or any of the last row of old ones cause things to jump around. I guess there is something about the new .col .s12 .m6 CSS that isn't playing well with the old CSS.
Happens in all browsers and I noticed it straight away. Let me know if you'd prefer a new ticket for that.

jmcglone:
@conorom the logo movement on hover has been that way for a while, it wasn't introduced with these new logos. If you check production now you'll see the middle logo on the first row moves up a bit and the NYU logo in the last row does as well. Better to make a separate ticket to address.

PR #2872, opened 2020-11-03
HELIO-3526 COUNTER reports for "ALL" institutions should be emailed v…

sethaj:
…ia background job

This works (tested on staging) and is self-contained so is safe to merge. However, I don't know if some of the language used needs to be approved by anyone (screenshots in the ticket). Probably just merge it and we can sort out specific wording later since it's easy enough to change.


sethaj:
We can't use gzip, need to use zip. Closing this PR to make that change.

PR #2890, opened 2020-12-11
Upgrade Skylight gem to 5.0beta

gitKrystan:
Thank you so much for participating in the [Skylight for Open Source](https://www.skylight.io/oss) program. We'd like to invite Fulcrum to participate in the public alpha for Source Locations for Skylight!

With this new feature, Skylight can help you pinpoint the locations in your code that correspond to events in the Skylight Event Sequence. This feature will allow your contributors to more easily find out exactly where an expensive operation originated without having to scour your code.

When you merge this PR, we will enable the feature flag for you on Skylight's servers. Then, once you deploy, you should begin seeing source locations data shortly after.

You can find documentation about the new feature at [https://www.skylight.io/support/source-locations](https://www.skylight.io/support/source-locations), and as always, if you have any questions or feedback, please get in touch with us either by responding here, emailing support@skylight.io, or via the in-app communicator at the lower-right of your Skylight dashboard.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -188,7 +188,7 @@ gem 'oauth2', '~> 1.2'
 gem 'signet'
 
 # performance profiling
-gem 'skylight'
+gem 'skylight', '~> 5.0.0.beta4'
 
 # Rails Webpack Tool
 gem 'webpacker', '~> 5.2.1'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -931,10 +931,8 @@ GEM
       tilt (~> 2.0)
     sitemap_generator (6.0.1)
       builder (~> 3.0)
-    skylight (3.1.1)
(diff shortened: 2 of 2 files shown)


gitKrystan:
Thanks for merging! I enabled the feature for your app, so you should see source locations on your endpoint pages now.

image
 from https://www.skylight.io/app/applications/TtaAmIlZOoFS/1611853200/31m/endpoints/MonographCatalogController%23index?responseType=html (I don't think source locations are visible on the public version of your Skylight dashboard just yet, so you'll need to log in to see them.)

Also worth noting is that the source locations feature is more useful if you also enable Skylight deploy tracking. You can learn how to do so [here](https://www.skylight.io/support/advanced-setup#deploy-tracking). With deploy tracking enabled, you'd be able to see source locations per deploy to determine if the source code changed between deploys.

PR #2897, opened 2021-01-06
HELIO-3650 Add circle ci

sethaj:
All specs are passing. We should probably merge this and run both this and travis for a while and then decide which one to keep.


sethaj:
Still need to make sure coveralls is working. Closing for now.

PR #2941, opened 2021-03-03
sighrax refactor

gkostin1966:
This one


app/models/sighrax/entity.rb
```diff
@@ -58,21 +47,19 @@ def type
...
+      def scalar(key)
+        vector(key).first
       end
-  end
 
-  class NullEntity < Entity
-    private_class_method :new
+      def vector(key)
+        Array(data[key])
+      end
```
botimer, line 56: I love these. The usage reads much tidier to me now!

PR #2946, opened 2021-03-05
sighrax patch

gkostin1966:
Refactoring (renaming) of class methods to better align with the domain model.


botimer:
This snuck by me! I love that the domain language is starting to poke through in more consistent and obvious ways. 😍

PR #2983, opened 2021-04-05
enable OA/free survey in PDF reader

conorom:
HELIO-3838


conorom:
dammit this was meant to be a hot fix. Thanks @sethaj

PR #3000, opened 2021-04-30
Helio 3810/pdf metadata watermark

conorom:
HELIO-3810
HELIO-3812


conorom:
Closing this to verify wording and to also make a change relating to Unicode characters in the PDF metadata.

PR #3012, opened 2021-05-24
replace every-page watermark with cover page

conorom:
HELIO-3812


conorom:
Closing after today's stand-up decision to add Copyright Holder to the cover page only if it is present (relates to https://tools.lib.umich.edu/jira/browse/HELIO-3919)

PR #3164, opened 2021-12-01
HELIO-4122 Filter crawlers from COUNTER stats

sethaj:


conorom:
@sethaj This is merged but isn't it weird that the institution ids are strings in the line:
`return if (@controller.current_institutions.map(&:identifier) & ["490", "2405", "2334", "2402"]).present?`?
Following the code down a bit I guess it’s cause they are never converted to int from Keycard request attributes, or something like that. It’s kind of extra confusing because institution id’s in our DB are integers, of course.
Stupid thing to start focusing on, I know.

PR #3165, opened 2021-12-02
Resolves CSB-261 Add aria-label to TOC buttons

gkostin1966:


gkostin1966:
Replicates button text, going to update the ticket instead.

PR #3198, opened 2022-01-27
stop auto-adjusting height of audio embed iframes

conorom:
HELIO-4167 (part 1)


sethaj:
This is release v2.87.1 and is in production.

PR #3226, opened 2022-03-07
HELIO-4143,  HELIO-3778  blacklight oai and irus_analytics counter

sethaj:


spec/requests/wayfless_spec.rb
```diff
@@ -73,6 +75,7 @@
       cover.save!
       epub.save!
       fr
+      allow_any_instance_of(IrusAnalytics::Controller::AnalyticsBehaviour).to receive(:send_irus_analytics_investigation).and_return(true)
```
gkostin1966, line 78: Is this redundant given irus has been stubbed out in the before block?

sethaj: Yeah, thanks. Pretty sure that was there when I was trying different test/mock strategies, I'll get rid of it.

PR #3318, opened 2022-07-22
non-modal panel buttons also close panels

conorom:
HELIO-4287
CSB PR already merged


app/views/epub_ebooks/show.html.erb
```diff
@@ -147,9 +147,6 @@
         })
         <% end %>
 
-        // No modal panel initiate
-        <%= render "cozy_controls_nomodal", presenter: @presenter, parent_presenter: @parent_presenter %>
```
jmcglone, line 151: @conorom I believe the code in here is still needed...there are some things that could be removed in the file, but I'm pretty certain lines 1-18 of `app/views/epub_ebooks/_cozy_controls_nomodal.html.erb` still need to be executed.

jmcglone:
Browser testing locally and on staging revealed this does not work in FF (102.0.1 (64-bit) but works in Chrome, Safari, and Edge. Further testing of CSB locally (w/o heliotrope) revealed FF is also an issue, so presumably this comes from https://github.com/mlibrary/cozy-sun-bear/pull/257/ and the fix is going to be in the CSB `too` branch. @conorom Want me to merge this and create a new CSB ticket for the FF issue?

conorom:
Thanks for catching that @jmcglone. You're right that the fix is in CSB. It's pretty minor so I'll just go ahead and merge another PR over there and then force-push the `yarn.lock` reference change in this one 🤠 🤠 🤠

PR #3358, opened 2022-10-28
generalize the automatic handle code

conorom:
HELIO-4327

### DIFF
--- a/app/controllers/embed_controller.rb
+++ b/app/controllers/embed_controller.rb
@@ -2,11 +2,11 @@
 
 class EmbedController < ApplicationController
   include IrusAnalytics::Controller::AnalyticsBehaviour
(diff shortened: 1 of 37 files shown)


spec/requests/api/opds/V2/feeds_spec.rb
```diff
@@ -125,6 +125,7 @@
 
         it 'is non-empty' do
           get api_opds_amherst_path, headers: headers
+          pp response.body
```
sethaj, line 128: Probably don't want the pp in here.

conorom: Thanks bud. This spec being broken really had me stumped. It was eating the error and this pp really did the job!!
Will remove.

PR #3390, opened 2022-12-12
HELIO-4226 force utf8 encoding for file_set.label in CharacterizeJob

sethaj:


conorom:
As discussed in Sprint Release Prep 146 today, we'll leave this open until I fully review why I can get this (very strange) problem to happen and Seth can't. It's low priority anyway, will get back to it in January.

PR #3401, opened 2023-01-06
HELIO-4391, HELIO-4393 (article creator field and spiel)

conorom:
Two commits for these tickets, respectively:
HELIO-4391
HELIO-4393

Code going on preview now.

### DIFF
--- a/app/presenters/hyrax/file_set_presenter.rb
+++ b/app/presenters/hyrax/file_set_presenter.rb
@@ -102,6 +102,25 @@ module Hyrax
       end
     end
 
+    def unreverse_names(comma_separated_names)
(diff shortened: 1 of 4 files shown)


conorom:
Closing this combo PR in favor of separating so that each ticket has its own PR.
Simpler in the end, especially as the first commit needs an importer fix not caught by the spec: using `multivalued: :yes_multiline` on [this line](https://github.com/mlibrary/heliotrope/blob/97575b3365dd9d2aaf6e9a9b992572ebb9f9c4a5/lib/metadata_fields.rb#L74). Thanks for spotting that @sethaj

PR #3426, opened 2023-02-07
HELIO-4407 add needs_kbart and group_key to product

sethaj:
Should probably have @jgmorse verify that adding these fields won't cause any problems for greensub


jgmorse:
If the fields aren't required and the API is unchanged, then greensub would be unaffected.

PR #3431, opened 2023-02-14
HELIO-4408 table to hold kbart info and monograph after_save action

sethaj:


sethaj:
We're going to take a different approach with this.

PR #3496, opened 2023-07-26
HELIO-4493 - extract multiple Monographs; add dir/NOID checks

conorom:
HELIO-4493


tbelc:
These updates look good. Thanks.

PR #3647, opened 2024-05-23
HELIO-4640 - update pdf.js to 3.11.174

conorom:
HELIO-4640


conorom:
@jmcglone this is on preview for your perusal

PR #3672, opened 2024-07-30
HELIO-4704 Attempt to speed up sitemap generation

sethaj:
It's maybe 2x faster now which is ok.

### DIFF
--- a/config/sitemap.rb
+++ b/config/sitemap.rb
@@ -36,28 +36,44 @@ SitemapGenerator::Sitemap.create(compress: false, create_index: true, max_sitema
   # add '/static/about_daily', changefreq: 'monthly'
   # add '/static/about_project', changefreq: 'monthly'
   # add '/static/rights', changefreq: 'monthly'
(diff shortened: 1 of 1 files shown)


config/sitemap.rb
```diff
@@ -36,28 +36,44 @@
...
+    file_set_ids = doc['file_set_ids_ssim']
+    cover_id = doc['representative_id_ssim']&.first
+    rep_file_set_ids = FeaturedRepresentative.where(work_id: doc['id'], file_set_id: file_set_ids).pluck(:file_set_id)
 
-    cover_id = d['representative_id_ssim']&.first
+    file_set_docs = []
+    # We have books with so many file_sets that the query is too big for solr, so we break them up to manage query size if needed
+    if file_set_ids.present?
+      until file_set_ids.empty?
+        file_set_docs << ActiveFedora::SolrService.query("{!terms f=id}#{file_set_ids.shift(999).join(",")}", fl: ['id', 'visibility_ssi'], rows: 1000)
```
conorom, line 58: @sethaj I much prefer getting all of a Monograph's FileSet docs using a single query like [this](https://github.com/mlibrary/heliotrope/blob/30ccfb347e5be7156abcfec2538be64ad026c0bc/lib/tasks/backup_monographs.rake#L86).

PR #3682, opened 2024-09-03
HELIO-4712 HELIO-4714 - scholarlyiq counter urls, blanks; final pre-SiQ launch data changes

conorom:
HELIO-4712
HELIO-4714


conorom:
@sethaj Put a tiny second commit on this because write_headers was a nonsense addition to the CSV file creation!!
It doesn't do anything without adding a headers parameter [[1](https://ruby-doc.org/core-3.0.0/doc/csv/options/generating/write_headers_rdoc.html)] [[2](https://stackoverflow.com/a/15912856)] and, well, these tasks all add a header row manually.

PR #3715, opened 2024-11-05
HELIO-4569 - no admin ebook download on public pages

conorom:
HELIO-4569


conorom:
@sethaj even with all the comments I feel this one is going to be tough to follow!
Basically:
1.  removing the ebook download button on the Monograph catalog page for admins if a public user would not see it
2. still allowing all the downloads on the show/manage page for admins
3. opening the door to doing the same as 1 for the FileSet show page download button

The biggest thing is that HELIO-4568 does not happen again and I've got a spec that checks that in here 🤞

PR #3986, opened 2026-05-15
HELIO-5072 chop obnoxiously long UA strings for COUNTER session

sethaj:


sethaj:
We're going to do this as a hotfix so I've changed the branch name it branched it off of tag 4.37. I'll open this as a different PR after release.

